PR #1, opened 2017-05-12
Use Compat rather than custom logic

ararslan:
Also added precompilation and made the Julia version check more specific.


quinnj:
Looks like we need to import the `xor` operator for the tests on 0.5

PR #2, opened 2017-05-12
More definitions

ararslan:
A wee bit o' PR scope creep. I was just going to add `identity`, then suddenly...

* Export `Null` to make it easier for other people to define methods
* Add iteration functions `start`, `next`, and `done`
* Remove unary `*` and `/` (DataArrays had this and it makes no sense)
* Add `identity`, `isnan`, `iszero`
* Add more tests

### DIFF
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -4,9 +4,9 @@ module Nulls
 
 using Compat
 importall Base.Operators
-import Compat: xor
+import Compat: xor, iszero
 
-export null, ?
+export null, Null, ?
 
 if VERSION < v"0.6.0-dev.2746"
     immutable Null end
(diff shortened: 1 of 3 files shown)


src/Nulls.jl
```diff
@@ -42,19 +47,18 @@ Base.isless(::Null, b) = false
...
           :(Base.log2), :(Base.exponent), :(Base.sqrt), :(Base.gamma), :(Base.lgamma),
-          :(Base.ceil), :(Base.floor), :(Base.round), :(Base.trunc)
-         )
+          :(Base.ceil), :(Base.floor), :(Base.round), :(Base.trunc))
     @eval $(f)(d::Null) = null
 end
 
 for f in (:(Base.iseven), :(Base.ispow2), :(Base.isfinite), :(Base.isinf), :(Base.isodd),
-          :(Base.isinteger), :(Base.isreal), :(Base.isempty))
+          :(Base.isinteger), :(Base.isreal), :(Base.isempty), :(Base.isnan), :(Base.iszero))
```
quinnj: Looks like we need `isimag` here

quinnj: And `iszero` isn't defined for 0.5. Is there `Compat` for that?

ararslan: Yep, just realized that. Also realized that `iszero` is a 0.6 thing, not 0.5, so I removed that.

ararslan: Oh hm, not sure re Compat. Will check.

ararslan:
If we ever rethink equality of `null` with other types returning a `Bool`, we should change `iszero`.

src/Nulls.jl
ararslan: Yes, it's in Compat. Good call!

ararslan:
Dangit, I thought I turned off Codecov comments! >:(

ararslan:
Hah, I'm pretty sure the 69% coverage decrease isn't real, especially given that I _added_ tests.

PR #8, opened 2017-05-17
Add convert(::Type{Null}, ::Null) method

nalimilan:
Fixes an ambiguity. Also make an existing test stricter.

@quinnj What do you need `convert(::Type{Null}, x)` for? Looks like a possibly dangerous method.

### DIFF
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -34,6 +34,7 @@ Base.done(::Null, b::Bool) = b
 
 Base.promote_rule{T}(::Type{T}, ::Type{Null}) = Union{T, Null}
 Base.convert(::Type{Null}, x) = null
+Base.convert(::Type{Null}, ::Null) = null
 
 # Comparison operators
 ==(::Null, ::Null) = true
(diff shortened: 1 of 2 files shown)


davidanthoff:
Ha, thanks, I had just added this locally as well and was about to open a PR :)

ararslan:
@staticfloat, looks like something is wrong with AppVeyor?

quinnj:
Yeah, I tried to copy/paste the appveyor changes from everywhere else, but maybe I messed something up?

nalimilan:
Looks like this line is lacking (at least):
```
  - ps: "[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.Securi
tyProtocolType]::Tls12"
```

ararslan:
Fixed in #9

nalimilan:
@quinnj Do you really think we need `convert(::Type{Null}, ::Any)`? In general I don't like the idea that you could transform valid values into nulls without notice. Even R prints a warning in these cases. This problem could happen for example if an `Array{Union{T, Null}}` ends up being an `Array{Null}` because the data contained only nulls, and you try to assign a value to it. Better throw a conversion error in that case.

quinnj:
No, we can take it out. I think I ran across something in transition and decided to add it; on the one hand, it makes sense "anything can be a null!", but I recognize it probably leads to more unintentional behavior than is worth it.

PR #12, opened 2017-05-20
Disallow conversion to null

nalimilan:
This can be dangerous since it allows creating missing values from valid ones,
which can happen e.g. when assigning to an `Array{Null}` which was expected to be
`Array{Union{T, Null}}`.

As discussed in https://github.com/JuliaData/Nulls.jl/pull/8.

### DIFF
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -45,9 +45,6 @@ Base.next(::Null, ::Bool) = (null, true)
 Base.done(::Null, b::Bool) = b
 
 Base.promote_rule{T}(::Type{T}, ::Type{Null}) = Union{T, Null}
-Null(x) = null
-Base.convert(::Type{Null}, ::Null) = null
-Base.convert{T}(::Type{Union{T, Null}}, x) = convert(T, x)
 
 # Comparison operators
 ==(::Null, ::Null) = true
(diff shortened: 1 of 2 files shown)


src/Nulls.jl
```diff
@@ -45,9 +45,6 @@ Base.next(::Null, ::Bool) = (null, true)
 Base.done(::Null, b::Bool) = b
 
 Base.promote_rule{T}(::Type{T}, ::Type{Null}) = Union{T, Null}
-Null(x) = null
-Base.convert(::Type{Null}, ::Null) = null
-Base.convert{T}(::Type{Union{T, Null}}, x) = convert(T, x)
```
quinnj: Can we keep this one? Otherwise, I see errors like
```julia
LoadError: MethodError: Cannot `convert` an object of type Float64 to an object of type Union{Int64, Nulls.Null}
This may have arisen from a call to the constructor Union{Int64, Nulls.Null}(...),
since type constructors fall back to convert methods.
```
when trying to convert a `Vector{?Float64}([1.0, 2.0])` to a `Vector{?Int}`

nalimilan: Oops, I was too enthusiastic. See https://github.com/JuliaData/Nulls.jl/pull/14.

PR #15, opened 2017-05-24
Add dropnull & dropnull! methods

cjprybol:


ararslan:
Seems like instead of introducing `dropnull` we should just use `filter(!isnull, ...)`, which is what this is calling anyway.

quinnj:
We also already have `Nulls.skip(A)` and `Nulls.replace(A, b)` which are convenience methods that produce generators. Just calling `collect` on either one should give us all the functionality we need.

nalimilan:
I wonder how we could provide a nice API for working with data frames/table and the split-apply-combine strategy. For example, this very common command:
```juli
aggregate(iris, :Species, mean)
```
is useless for any dataset with missing data since it will return `null`. Yet the null-skipping form is annoyingly long:
```julia
aggregate(iris, :Species, x -> sum(Nulls.skip(x)))
```

Maybe we could support this?
```julia
aggregate(iris, :Species, Nulls.skip(sum))
```
(i.e. `Nulls.skip(f::Function) = x -> f(Nulls.skip(x))`)

I would even advocate for skipping nulls by default in all `DataTable` functions. Possibly via a property of the `DataTable` which would only need to be set once for all. (This is of course out of scope in this repo, but is important to take into account for the design.)

PR #16, opened 2017-05-24
Extend StatsBase.describe to better handle Nulls

cjprybol:
https://github.com/JuliaData/DataTables.jl/pull/66#discussion_r117849476


quinnj:
This is great; I kind of want to put this in DataTables for now, since it already has StatsBase as a dependency. Thoughts?

cjprybol:
Sounds good! I'll open another PR over there.

PR #24, opened 2017-08-24
Remove ?T print style

cjprybol:
https://github.com/JuliaData/Nulls.jl/commit/b1a555d119a7537abe6e8f57451c19f7caeece46##commitcomment-23842283 and https://github.com/JuliaData/DataTables.jl/pull/83#discussion_r134702649

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -131,5 +131,4 @@ using Base.Test, Nulls
     @test Nulls.T(Any) == Any
     io = IOBuffer()
     show(io, Union{Float64, Null})
-    @test String(take!(io)) == "?Float64"
 end
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -10,8 +10,6 @@ struct Null end
 const null = Null()
 
 Base.show(io::IO, x::Null) = print(io, "null")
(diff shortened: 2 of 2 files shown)


test/runtests.jl
```diff
@@ -131,5 +131,5 @@ using Base.Test, Nulls
     @test Nulls.T(Any) == Any
     io = IOBuffer()
     show(io, Union{Float64, Null})
-    @test String(take!(io)) == "?Float64"
+    @test String(take!(io)) == "Union{Nulls.Null, Float64}"
```
quinnj: Don't worry about that; just make the test say `@test String(take!(io)) in ("Union{Nulls.Null, Float64}", "Union{Float64, Nulls.Null}")`. There's a ton of logic that goes into the Union type sorting that's in 0.7 only.

ararslan: You know it'll be one order or the other regardless of the version, so I'd do
```julia
@test String(take!(io)) in ["Union{Nulls.Null, Float64}", "Union{Float64, Nulls.Null}"]
```

ararslan: Jacob ninja'd my recommendation

quinnj: Actually. just take this test out for now; it's not like we're actually testing anything here anyway.

PR #27, opened 2017-08-29
Remove show method already handled by Base.show

cjprybol:
I don't think we need this, seems to be handled fine by Base.show.

If this passes I'll merge and tag a new 0.0.6 release, which we will need to set as the new version minimum for DataTables and CategoricalArrays which will stop/have stopped using the `?T` syntax present in release 0.0.5

### DIFF
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -10,7 +10,6 @@ struct Null end
 const null = Null()
 
 Base.show(io::IO, x::Null) = print(io, "null")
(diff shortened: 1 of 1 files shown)


quinnj:
Yeah, we only needed this because of the conflict w/ our custom show method that we already took out.

quinnj:
No use waiting on osx travis; everything else is fine.

cjprybol:
Thanks Jacob!

PR #28, opened 2017-08-30
Add coalesce() function

nalimilan:
Small helper function which is useful either to take the first non-missing
value, or to replace missing values with a given value (when passing a value known
not to be missing). Also usable in its broadcasted form, where a non-null
scalar argument will be automatically used in place of all nulls.
Inspired by the equivalent SQL and dplyr function.

Currently, broadcast returns a `Vector{Any}` rather than the more efficient
`Vector{Union{Null, T}}` when the result contains nulls. This will have to
be improved in Base.

The function is not exported (like `replace()` and `skip()`), but it could be
at some point since its name is unlikely to conflict with other packages.

### DIFF
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -96,4 +96,41 @@ replace(itr, a, b) = (ifelse(v == a, b, v) for v in itr)
 replace(itr, b) = replace(itr, null, b)
 skip(itr, a=null) = (v for v in itr if v != a)
 
+"""
+    coalesce(x, y...)
+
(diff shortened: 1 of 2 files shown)


quinnj:
Nice, I like it.

nalimilan:
Wow, fast! One issue I just realized is that this function operates on scalar, while `replace()` and `skip()` operate on iterators. Maybe not a big deal, but...

quinnj:
I think that should be fine; I think all 3 methods can have their place in different workflows.

PR #31, opened 2017-09-27
Add zero() methods

nalimilan:
Provided by DataArrays.

### DIFF
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -50,7 +50,7 @@ if VERSION < v"0.7.0-DEV.300"
 end
 
 # Unary operators/functions
-for f in (:(+), :(-), :(Base.identity),
+for f in (:(+), :(-), :(Base.identity), :(Base.zero),
           :(Base.abs), :(Base.abs2), :(Base.sign),
           :(Base.acos), :(Base.acosh), :(Base.asin), :(Base.asinh), :(Base.atan), :(Base.atanh),
           :(Base.sin), :(Base.sinh), :(Base.cos), :(Base.cosh), :(Base.tan), :(Base.tanh),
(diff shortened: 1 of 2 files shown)


src/Nulls.jl
```diff
@@ -66,6 +66,8 @@ for f in (:(Base.iseven), :(Base.ispow2), :(Base.isfinite), :(Base.isinf), :(Bas
     @eval $(f)(d::Null) = false
 end
 
+Base.zero(::Type{Union{T, Null}}) where {T} = zero(T)
```
nalimilan: This method isn't great, as it turns `zero([])` into a `StackOverflowError` instead of a more explicit `MethodError`. But it's apparently needed to port DataArrays to Nulls, because it relies on it for reductions (by using code from Base).

ararslan: Given that
```julia
julia> a = Type{Union{T,N}} where {T,N}
Type{Union{T, N}} where N where T

julia> b = Union{Type{T},Type{N}} where {T,N}
Union{Type{T}, Type{N}} where N where T

julia> a == b
true
```
It seems that we could simply define `Base.zero(::Type{Null})` (even to be a `MethodError`) rather than specializing on `::Type{Union{T, Null}}`, and that would avoid a stack overflow. Is that true or am I missing something?

nalimilan: Unfortunately, that doesn't work for DataArrays, where `zero(Union{T, Null})` is called because that's the element type of the array on which the reduction is done. These issues with too broad signatures are quite common in Nulls.jl... :-/

ararslan: I'm still confused. Defining `Base.zero(::Type{Union{A, B}}) = c` should be the same as defining `zero(::Type{A}) = c` and `zero(::Type{B}) = c`. So even if we call `zero(eltype(x))` where `x` has element type `Union{T, Null}`, it seems we could provide just the specialization for `Null` and it would work as intended?

nalimilan: Hmm, no. How is the call to `zero(Union{Int, Null})` supposed to realize that `zero(Int) === zero(Null)`? Though we could imagine adding a recursive definition like that, which would call `zero` on each type of the union, and call `promote` on the resulting values.

ararslan: Ahhhhh okay, I see now. Thanks.

PR #32, opened 2017-09-27
Add Nulls.fail() iterator

nalimilan:
Provided by DataArrays as `each_nafail()`. It simply iterates over the
input and throws an error if a null is found.


ararslan:
I've actually never understood the purpose of that function in DataArrays. It seems equally easy (and much clearer) to do
```julia
any(isnull, x) && throw(NullException())
```
explicitly in user code as needed.

PR #41, opened 2017-10-04
Fix #23 by defining a few more helper promote_rule definitions

### LINKED ISSUE #23: Look into promotion issue
https://github.com/JuliaStats/DataArrays.jl/pull/280#discussion_r130244769

quinnj:
cc @andreasnoack @nalimilan @ararslan 

Fixes #23.

Any more tests we need here?

### DIFF
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -24,6 +24,11 @@ nulls(::Type{T}, dims...) where {T >: Null} = fill!(Array{T}(dims), null)
 nulls(::Type{T}, dims...) where {T} = fill!(Array{Union{T, Null}}(dims), null)
 
 Base.promote_rule(::Type{T}, ::Type{Null}) where {T} = Union{T, Null}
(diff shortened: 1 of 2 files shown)


nalimilan:
Looks good in technical terms, but can you remind me why we need these? When do we have to call `promote_type{Union{T, Null}, S)`? When promoting the eltypes of two arrays?

quinnj:
Yeah, otherwise you have situations like
```julia
julia> promote_type(Union{Null, Int}, Float64)
Any
```

nalimilan:
OK. Maybe this should be handled by Julia? I mean, we have `promote_type(::Null, ::Type{T}) where {T} = Union{Null, T)`, but maybe we should have more generally a fallback so that instead of returning `Any` when no promotion rule is defined, a `Union` should be returned instead? Like with `typejoin`, there would be some rules to decide when a `Union` would be too complex to be useful, so that  `Any` would be returned.

PR #42, opened 2017-10-05
Add rounding and transposition methods

nalimilan:
Used to be provided by DataArrays, needed for https://github.com/JuliaStats/DataArrays.jl/pull/288.


src/Nulls.jl
```diff
@@ -81,6 +80,15 @@ for f in (:(+), :(-), :(*), :(/), :(^),
     end
 end
 
+# Rounding and related functions
+for f in (:(Base.ceil), :(Base.floor), :(Base.round), :(Base.trunc))
+    @eval begin
+        ($f)(::Null, digits::Integer=0, base::Integer=0) = null
+        ($f)(::Type{>:Null}, ::Null) = null
```
nalimilan: This one is a bit special: I've considered `round(Int, null)` should throw a `NullException`, but `round(Union{Int, Null}, null)` should return `null`. That's safer than returning `null` whatever the requested type. When `NullException` supports error messages, we should use it to explain what to do.

quinnj: 👍

PR #44, opened 2017-10-12
Restrict zero function types

cjprybol:
Base only provides `zero(::Type{T}) where T<:Number`, not `zero(::Type{T}) where T`, and our method results in StackOverflows when called with `Any` when it should return an error. See https://github.com/JuliaData/DataFrames.jl/issues/1251

### DIFF
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -67,7 +67,7 @@ for f in (:(!), :(+), :(-), :(Base.identity), :(Base.zero),
     @eval $(f)(d::Null) = null
 end
 
-Base.zero(::Type{Union{T, Null}}) where {T} = zero(T)
+Base.zero(::Type{Union{T, Null}}) where {T <: Number} = zero(T)
 
 # Binary operators/functions
 for f in (:(+), :(-), :(*), :(/), :(^),


nalimilan:
@quinnj Please don't merge PRs during the night! Unless it's blocking something, I think a PR should sit for at least one day in case somebody has remarks.

We had already spotted the problem before introducing `zero` (https://github.com/JuliaData/Nulls.jl/pull/31/files#r141323811), but merged it anyway because `zero` doesn't not only work on `Number`. At the very least, it should also work for date types. Then it's debatable whether it's OK to add more methods for each type which might support `zero`, or whether we really need a generic fallback despite the `StackOverflow` error it triggers.

quinnj:
Sorry, I thought it had been open longer. I feel like we should be more conservative here. We can add a method for Union TimeTypes and more as needed, but I feel like it's too easy to bump into `zero` calls out in the wild to be triggering stack overflows.

nalimilan:
We should have a policy for this kind of situation. For example, we currently define `*` only for `Number`, so it doesn't work for strings. If we decide to overload functions on a case by case basis, we should add `AbstractString` methods. The other approach is to accept `Any` arguments.

nalimilan:
See https://github.com/JuliaData/Nulls.jl/pull/45.

PR #46, opened 2017-10-12
Add levels() function

nalimilan:
Nulls is a natural home for this method since it skips null values,
and it needs to be defined in an essential package so that both DataArrays
and CategoricalArrays can override it, and other packages use it without
conflicts nor heavy dependencies.

### DIFF
--- a/src/Nulls.jl
+++ b/src/Nulls.jl
@@ -4,7 +4,7 @@ module Nulls
 import Base: *, <, ==, !=, <=, !, +, -, ^, /, &, |, xor
 using Compat: AbstractRange
 
-export null, nulls, Null
+export null, nulls, Null, levels
 
 struct Null end
 
@@ -149,6 +149,25 @@ julia> coalesce.([null, 1, null], [0, 10, 5])
 """
 coalesce(x) = x
 coalesce(x, y...) = ifelse(x !== null, x, coalesce(y...))
+
+"""
+    levels(x)
+
+Return a vector of unique values which occur or could occur in collection `x`,
+omitting `null` even if present. Values are returned in the preferred order
+for the collection, with the result of [`sort`](@ref) as a default.
+
+Contrary to [`unique`](@ref), this function may return values which do not
+actually occur in the data, and does not preserve their order of appearance in `x`.
+"""
+function levels(x)
+    T = Nulls.T(eltype(x))
+    levs = convert(AbstractArray{T}, filter!(!isnull, unique(x)))
+    if method_exists(isless, Tuple{T, T})
+        try; sort!(levs); end
+    end
+    levs
+end
 
 # AbstractArray{>:Null} functions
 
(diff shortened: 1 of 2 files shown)


src/Nulls.jl
```diff
@@ -150,6 +150,25 @@ julia> coalesce.([null, 1, null], [0, 10, 5])
...
+Return a vector of unique values which occur or could occur in collection `x`,
+omitting `null` even if present. Values are returned in the preferred order
+for the collection, with the result of [`sort`](@ref) as a default.
+
+Contrary to [`unique`](@ref), this function may return values which do not
+actually occur in the data, and does not preserve their order of appearance in `x`.
+"""
+function levels(x)
+    T = Nulls.T(eltype(x))
+    levs = convert(AbstractArray{T}, filter(!isnull, unique(x)))
```
ararslan: `unique` should never alias, so this might be more efficient with `filter!`.

nalimilan: I would have sweared I had used `filter!`. Too many `!` in a row...

ararslan:
Seems weird to me that something like this would be defined in Nulls, but I don't have a better suggestion, so...

nalimilan:
Define it in Base? :-p

BTW, cf. discussion at https://github.com/JuliaLang/julia/pull/20800.

src/Nulls.jl
```diff
@@ -150,6 +150,25 @@ julia> coalesce.([null, 1, null], [0, 10, 5])
...
+    levels(x)
+
+Return a vector of unique values which occur or could occur in collection `x`,
+omitting `null` even if present. Values are returned in the preferred order
+for the collection, with the result of [`sort`](@ref) as a default.
+
+Contrary to [`unique`](@ref), this function may return values which do not
+actually occur in the data, and does not preserve their order of appearance in `x`.
+"""
+function levels(x)
```
quinnj: should we restrict the argument to `AbstractArray`?

nalimilan: I've considered it, but I saw no reason to reject iterables since `unique` supports them.

quinnj:
Anything we can do about the nightly failure?

nalimilan:
Looks like the same Julia bug which has plagued DataFrames (or a related package, I don't remember) for several weeks?

nalimilan:
Filed https://github.com/JuliaLang/julia/issues/24124.

src/Nulls.jl
ararslan: One argument in favor of allowing iterables is that `levels(Nulls.skip(x))` wouldn't work otherwise.

nalimilan: Since `levels` skip nulls, you shouldn't need to skip them. But yeah, if somebody passed your function an iterator in order to skip nulls and you call `levels` on that, it's nice that it works.

nalimilan:
OK to merge?

PR #49, opened 2017-10-14
Add float(::AbstractArray{Union{T, Null}}) method

nalimilan:


src/Nulls.jl
```diff
@@ -211,4 +211,11 @@ function Base.all(f, A::AbstractArray{>:Null})
     return anynull ? null : true
 end
 
+function Base.float(A::AbstractArray{Union{T, Null}}) where {T}
+    U = typeof(float(zero(T)))
+    convert(AbstractArray{Union{U, Null}}, A)
+end
+
+Base.float(A::AbstractArray{Null}) = throw(NullException())
```
quinnj: Hmmmm....should we define the scalar `float` as well? And in that case, wouldn't we want `Base.float(A::AbstractArray{Null}) = A`? (since the scalar would be `Base.float(x) = null`. Is that too liberal/dangerous do you think?

nalimilan: Makes sense. `float` is already quite liberal since it works e.g. on `Complex`, so I'd say it's fine to allow nulls. I've added the scalar method and updated the vectorized method to accept arrays of nulls.

PR #51, opened 2017-11-12
Rename to Missing and missing

nalimilan:
Stop using `Base.isnull` and `Base.NullException`. This allows adding a `msg` field to the new `MissingException`.

This is mostly a systematic replacement of `null` with `missing` and so on. I suggest we rename the repo to Missings.jl afterwards, so that we keep the history and avoid creating yet another package for missing values. The package will be used until (if?) the features are included in Julia, and in particular on Julia 0.6.

We should discuss and adapt later functions like `skip`, `fail` and `replace` if we want to change them.


nalimilan:
I've added a commit to use `⯑` (uncertainty character) for compact printing, but we can discuss using other symbols.

src/Missings.jl
```diff
@@ -4,13 +4,38 @@ module Missings
...
+"""
 const missing = Missing()
 
-Base.show(io::IO, x::Missing) = print(io, "missing")
+"""
+    ⯑
+
+Compact representation of [`missing`](@ref), denoting a missing value.
+"""
+const ⯑ = missing
```
ararslan: Wait, for real?

ararslan:
A problem with `⯑` is that some fonts use that for characters they can't represent, like subscripts.

nalimilan:
> A problem with ⯑ is that some fonts use that for characters they can't represent, like subscripts.

AFAIK that's `�`. If we're afraid of the confusion, we can always use another character, though.

src/Missings.jl
nalimilan: Yes, the idea being that when people see that character, they can just copy/paste it to the REPL and see what it means. That's much better than the pseudo-values we printed with `#NULL`.

ararslan:
screen shot 2017-11-13 at 10 47 51 am

alyst:
I have to admit, `missing` is 75% longer than `null` to type.
I appreciate that missing values are respected so much, but it's not very practical to type `missing` or `ismissing` every time.
Also plural form "Missings" is a little bit awkward, yet it would be used a lot (`Missings.skip()`).
Maybe it makes sense to reintroduce good old `NA` as a synonym/rename? `NAs.jl` sounds good to me, I hope nobody is going to implement network storage in embedded Julia.

nalimilan:
Sorry, that's too late now, you should have come to #data to discuss it. Note that with autocompletion, you just need to type `mis[tab]`, so it's not too bad. Regarding `Missings.skip`, we're looking for a shorter replacement. Since these features are going to be included in Base, the function wouldn't need to be qualified.

alyst:
> Sorry, that's too late now, you should have come to #data to discuss it.

Indeed, that's my fault. I have 2 weak excuses: 1) last week I was at the place with abysmal internet connection, 2) I don't know what is `#data`; `README.md` doesn't mention it, neither does `README.md` and `CONTRIBUTING.md` of `DataFrames.jl`. Maybe in the future for changes like this (renaming the package, although it was used only by the master versions of `JuliaData` packages) there should be a short deprecation period (2-4 weeks) to notify the early adopters and collect a broader feedback.

> Note that with autocompletion, you just need to type mis[tab], so it's not too bad.

That might be true for some autocompletion implementations and in the absence of ambiguities. However, I'm rather concerned about resulting readability of the code. Especially if `Union{T, Missing}` is not gonna be replaced by `T?`, it would definitely bloat up type declarations of function arguments as well as dispatch error messages (which are already quite hard to read for things like `SubArray{CategoricalArray}`).

cjprybol:
By #data Milan is referring to the data channel on the Julia slack account. After weeks of discussion (in addition to the years of discussion on various GitHub issues), a consensus arose that the distinction between `nothing` and `null` was unclear depending on context, but everyone seemed to agree that the distinction between `nothing` and `missing` was very clear. Reasons for not choosing `NA` have been pretty consistent throughout the discussions. That's the best brief catch-up summary I can give, but if you'd like to read or discuss more, the instructions for joining slack are on the [community page](https://julialang.org/community/). As I understand it, `missing` is the only terminology that was deemed clear enough to include in base, which was another large factor in the renaming since everyone seems to want this functionality in base even though there will never be unanimous agreement on what to name it.

alyst:
I'm aware of the long nothing/missing/NA/null discussions. I thought `Nulls` was already the consensus. It was surprising to see that after the recent introduction of `Nulls.jl` it's renamed without prior notice in README.md and/or deprecation warnings given that JuliaData/DataFrames.jl#1232 is halfway through.

davidanthoff:
As far as I can tell, the idea to rename ``null`` to ``missing`` was first suggested on slack on 11/9 and then merged yesterday, five days later with no discussion anywhere else. That doesn’t strike me as an ideal process for such a big decision. I think it would be a good idea to a) have such discussions on github, where I think more people are reached and b) give them a bit more time.

mkborregaard:
As an extra aside, Slack does not archive conversations, so the entire discussion is lost for eternity. This means that there sool will exist no record of the arguments that went into the decision.

nalimilan:
> It was surprising to see that after the recent introduction of Nulls.jl it's renamed without prior notice in README.md and/or deprecation warnings given that JuliaData/DataFrames.jl#1232 is halfway through.

"Halfway through" is clearly an overstatement... :-)

Sorry if this sounds too sudden, but `missing` got a (surprisingly) broad support on Slack, and I honestly don't think we can really have a technical discussion to make this choice. In the end it's really subjective, so given that @StefanKarpinski proposed `missing`, that JuliaData people supported it and than it received a large majority of support on Slack, I think it's not unreasonable to consider it OK to merge this PR after 4 days without any objection. It's too bad this was precisely the period when you were offline, but I wasn't aware of that. The goal was really to make progress on the missing values front so that we can release the long-awaited DataFrames, which many people reclaim these days.

Note that the tagged releases of Nulls.jl continue to work, so nothing is broken. This PR is almost equivalent to creating a new package, except that we wanted to preserve the history to avoid the multiplication of missing values repos. Regarding the name, `T?` is still an open option, but it won't likely be in 1.0 because @StefanKarpinski preferred leaving some time to judge whether `Union{T, Missing}` is the best use for this syntax (rather than e.g. `Union{T, Void}`).

StefanKarpinski:
Here was my "impassioned" case for `missing` rather than `NA` or `null` on Slack:

---

given that every single person seems to explain `na` and `null` as “oh, that means this data point is missing”, I find it pretty hard to be convinced that we should call this anything but `missing`

it may be slightly annoying to change at this point, but this is a name we’re going to live with forever (presumably), so let’s just pick the right name now

the name should clearing indicate what it means, which is a property that only `missing` has of all the options we’ve discussed

I find it very hard to get clearer and more common sense than:
- `nothing` means “there is no value”
- `missing` means “there is a value but we don’t have it”

the fact that other systems have chosen manifestly bad, confusing names for this concept is not a good argument that we should also do so; especially given that there is no broad consensus:
* SQL and many other systems call this `null` or `NULL`
* C/C++/C#/Java use `null` to mean something else entirely
* R calls this `NA` (and uses `NULL` to mean something quite different)
* Pandas doesn’t have a consistent name for missing data
* STATA uses `missing` and displays it as `.`
* SAS mostly uses `.` but uses “missing” in function names
* SPSS displays as `.` but uses “missing” in function names (edited)

in short, how missing values are represented or handled in code is a compete and utter shitshow with almost no consistency across systems

my conclusion is that we should not attempt to mimic existing systems – because if so, then which one? – and this is our chance to do better

In addition to “missing” being much more self-explanatory, I think that the name “NA” has done real damage to R in that it has caused a lot of fear, uncertainty and doubt around what exactly NA means. There’s a lot of confusion about the distinction between invalid and missing values and which meaning you can or should use NA to represent. And this isn’t just an idle philosophical problem – there are statistical analyses where invalid data and missing data must be treated differently in order to come to the correct conclusion.

Using “NA” in Julia would inherit and perpetuate all of that confusion, whereas using “missing” is an explicit departure from R, not just in naming, but – as a consequence of that – also in clarity of meaning. We are saying that in Julia, `nothing` means that there is no value of the expected type whereas `missing` means that there *is* some value of the expected type but we don’t know what it is. These concepts are commonly conflated in R. The names “nothing” and “missing” reinforce the meaning and correct use of these values whereas “NA” since it has no obvious meaning (“not available”? “not applicable”?) allows fuzziness and lack of clarity.

StefanKarpinski:
The name `unknown` also came up but it was deemed possibly less clear and harder to type. In the future, it might be useful to distinguish yet another meaning: `invalid` which means that there was a value but it was invalid somehow. It's unclear how `invalid` would differ behaviorally from `nothing` but it may be useful to distinguish and we have the technology now to support small unions efficiently, so it would be entirely possible for an element type to be `Union{T, Nothing, Invalid, Missing}` at which point we have as rich semantic support for missing and invalid data as any system around. But if we choose `NA` then all of those (significant) distinctions seem like they will inevitably just get lumped into the fuzzy `NA` category.

nalimilan:
> n the future, it might be useful to distinguish yet another meaning: invalid which means that there was a value but it was invalid somehow.

I actually had similar thoughts recently. In R we indeed incorrectly conflate "not applicable"/`invalid` (e.g. last menstrual period for a male) with "not available"/`missing` (e.g. missing last menstrual period for a female). This has implications for the propagation of uncertainty, as `invalid == invalid` should return true, contrary to `missing == missing` which has to return `missing`. I agree that `nothing` could be enough, maybe we don't need a separate type. I've also considered having extended missing and invalid types carrying a payload to indicate the kind or reasons of invalidity/missingness, like e.g. the [FIHR](https://www.hl7.org/fhir/v3/NullFlavor/cs.html) does.

alyst:
@StefanKarpinski I'm totally with you: the distinction between null pointers, omitted and missing values is important; the current name selection is probably the best as long as the clarity is concerned. It will definitely help the newcomers to grasp the difference between the concepts and use them in a proper context.

I'm just trying to think how actual coding and reading code with `missing` would look like.
`missing` (and its derivatives `Missing`, `ismissing` etc) might be a little bit too long for such fundamental and frequently used concept.
Ideally, the moderately experienced coder should just "see through"/"type through" the syntax element representing missing value.
Yet `missing` sort of forces the person to read/write it and inflates the code overall.
Maybe there's nothing to do about, maybe it's still feasible to sacrifice a little bit of clarity for the newcomers and use a leaner term.
FWIW, `mis` (as long as there's no less fundamental `ans`)

> so it would be entirely possible for an element type to be `Union{T, Nothing, Invalid, Missing}` at which point we have as rich semantic support for missing and invalid data as any system around.

I'm a little bit scared to think how the code of e.g. `CategoricalArrays` supporting all possible variations of `T`, `Union{T, Missing}`, `Union{T, Invalid, Nothing}` etc as the element type would look like.

nalimilan:
Using `mis` would defeat the purpose of choosing an explicit term, what is more without the benefit of a term used in any other language. I don't think typing `missing` will be a big deal, and syntax highlighting will make it easy to read. The only place where it could be annoying IMHO is when printing arrays, which is why I proposed using a special symbol. We could also just print it in grey (where possible), generalizing what we currently do in DataFrames with `null`.

> I'm a little bit scared to think how the code of e.g. CategoricalArrays supporting all possible variations of T, Union{T, Missing}, Union{T, Invalid, Nothing} etc as the element type would look like.

I haven't thought it through yet, but at least it shouldn't be hard to support a single custom type of missing values (instead of hardcoding `Missing`) for a given `CategoricalArray`. From the start, I had in mind that we could use negative codes to indicate variants of missing values. Anyway better discuss this elsewhere.

PR #52, opened 2017-11-14
Remove remaining uses of ⯑

nalimilan:
I always forget that `export` does not throw an error when trying to export non-existent objects... :-/


nalimilan:
> Interesting that git lets you use that character in the branch name

Yes, that's so fun when you like spending time looking for a way to type the name to a branch on the keyboard!

PR #56, opened 2017-11-22
Addressing Breaking Change to Dates

Nosferican:
Addressing [0c0d6cc](https://github.com/JuliaLang/julia/commit/0c0d6cca5d3793eaf6321ef0577afd6ce594a9b5)
If Nulls is getting a version for Julia 0.7 it would need a similar fix (until [Compat](https://github.com/JuliaLang/Compat.jl/pull/413)?).

### DIFF
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -2,6 +2,9 @@ __precompile__(true)
 module Missings
 
 import Base: *, <, ==, !=, <=, !, +, -, ^, /, &, |, xor
+if VERSION >= v"0.7.0-DEV.2575"
+    import Dates
+end
 using Compat: AbstractRange
 
 export ismissing, missing, missings, Missing, MissingException, levels, skipmissing
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -2,10 +2,7 @@ __precompile__(true)
 module Missings
 
 import Base: *, <, ==, !=, <=, !, +, -, ^, /, &, |, xor
-if VERSION >= v"0.7.0-DEV.2575"
-    import Dates
-end
-using Compat: AbstractRange
+using Compat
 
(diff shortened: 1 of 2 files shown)


src/Missings.jl
```diff
@@ -2,6 +2,9 @@ __precompile__(true)
 module Missings
 
 import Base: *, <, ==, !=, <=, !, +, -, ^, /, &, |, xor
+if VERSION >= v"0.7.0-DEV.2575"
+    import Dates
```
nalimilan: I guess `using` should be enough, right?

nalimilan:
Thanks. I think we'd better wait for Compat to support this, though.

src/Missings.jl
Nosferican: It should, but `import` is safer since `Base.Dates` used to export only certain names. I believe they should be the same now that it moved to stdlib.

nalimilan: No, it's not safer, as it allows overriding functions in `Dates`, which is not what we want to do AFAIK.

Nosferican: It depends on the style. `using` has no overriding allowed, but floods the namespace (especially with packages that export many functions and structs) and can cause issues if multiple packages have the same functions. `import` allows to override (though you still have to do it explicitly), but keeps the namespace clear. There is also a warning if any package overrides a method. I always use the module prefix (except with `Base`) since I saw it in an old Julia blog. Either one would do and probably Compat will probably check for the optimal solution.

omus:
https://github.com/JuliaLang/Compat.jl/pull/413

src/Missings.jl
nalimilan: `using` also allows specifying which objects you want to use.

nalimilan:
Thanks!

PR #57, opened 2017-11-24
Compat.Dates

Nosferican:
Added `Compat.Dates` for the main file and the test file.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,4 +1,4 @@
-using Base.Test, Missings, Compat.Dates
+using Base.Test, Missings
 
 @testset "Missings" begin
     # test promote rules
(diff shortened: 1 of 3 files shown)


Nosferican:
There were test errors on the nightly. Are those related?

nalimilan:
Yes, it seems. Can you avoid doing `using Compat.Dates`, and use `Compat.Dates.X` where needed (with `X` the type or method used)? That way it's clearer where we use Dates (in very few places actually).

test/runtests.jl
```diff
@@ -58,8 +58,8 @@ using Base.Test, Missings
     @test zero(Union{Int, Missing}) === 0
     @test zero(Union{Float64, Missing}) === 0.0
 
-    for T in (subtypes(Dates.DatePeriod)...,
-              subtypes(Dates.TimePeriod)...)
+    for T in (subtypes(Compat.Dates.DatePeriod)...,
```
nalimilan: Looks like you need `using Compat` at the top of the file.

Nosferican: Thanks.

PR #59, opened 2017-11-29
Support one and oneunit

omus:


src/Missings.jl
```diff
@@ -95,8 +95,9 @@ for f in (:(!), :(+), :(-), :(Base.identity), :(Base.zero),
     @eval $(f)(d::Missing) = missing
 end
 
-Base.zero(::Type{Union{T, Missing}}) where {T <: Number} = zero(T)
```
nalimilan: The type restriction was there on purpose. See https://github.com/JuliaData/Missings.jl/pull/44, and the workaround by defining `zero(::Type{Any})` [here](https://github.com/JuliaLang/julia/pull/24653/files#diff-c4a56019a8dae547ca5556ad8df4d1f0R63) (but this would be type piracy in this package).

omus: Thanks for pointing this out. I think I found a nice way to work around the type piracy problem.

omus:
> Feel free to push the same changes to my Julia PR.

Will update https://github.com/JuliaLang/julia/pull/24653

omus:
I believe this is ready to merge. Any other comments?

Issue #60, opened 2017-11-30
Shorthand for AbstractArray conversions


bkamins:
It would be nice to be have the following two functions:
1. takes `AbstractArray{T}` and returns `AbstractArray{Union{T, Missing}}`;
2. takes `AbstractArray{Union{T, Missing}}` and returns `AbstractArray{T}` or throws an error if it contained any `missing.

It now can be achieved e.g. using `convert` but I could not think of any more compact and handy conversion method (if there is such it would be great to document it).

If it is considered to be worth to add such methods there are two corner cases that is also currently present in `Missings.T` function:

```
julia> x = Union{Missing, Int, Float64}[1, 1.0, missing]
3-element Array{Union{Float64, Int64, Missings.Missing},1}:
 1
 1.0
  missing

julia> Missings.T(x)
ERROR: MethodError: no method matching T(::Array{Union{Float64, Int64, Missings.Missing},1})
Closest candidates are:
  T(::Type{Missings.Missing}) at D:\Software\JULIA_PKG\v0.6\Missings\src\Missings.jl:41
  T(::Type{Any}) at D:\Software\JULIA_PKG\v0.6\Missings\src\Missings.jl:43
  T(::Type{Union{Missings.Missing, T1}}) where T1 at D:\Software\JULIA_PKG\v0.6\Missings\src\Missings.jl:40
  ...
```
This should probably return `Union{Int, Float64}`.

The other is:
```
julia> Missings.T([1,2,3])
ERROR: MethodError: no method matching T(::Array{Int64,1})
Closest candidates are:
  T(::Type{Missings.Missing}) at D:\Software\JULIA_PKG\v0.6\Missings\src\Missings.jl:41
  T(::Type{Any}) at D:\Software\JULIA_PKG\v0.6\Missings\src\Missings.jl:43
  T(::Type{Union{Missings.Missing, T1}}) where T1 at D:\Software\JULIA_PKG\v0.6\Missings\src\Missings.jl:40
  ...
```
which should probably be `Int` and not error.

ararslan:
Your 1. there is `allowmissing!`, though that might be in DataFrames rather than here, I can't recall. 2. is dealt with using an iterator that fails on `missing` during iteration.

`Missings.T` is intended to act on types rather than values, so I don't think we should add other methods to it. It's possible to make it work for larger unions with something like this:
```julia
Missings.T(U::Type{Union{S,Missing}}) where {S} =
[...]

PR #64, opened 2017-12-02
Add allowmissing() and disallowmissing() convenience functions

nalimilan:
As requested by several people, given that the convert() call is complex
and not easy to discover.

Fixes https://github.com/JuliaData/Missings.jl/issues/60.

### DIFF
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -4,7 +4,8 @@ module Missings
 import Base: *, <, ==, !=, <=, !, +, -, ^, /, &, |, xor
 using Compat
 
-export ismissing, missing, missings, Missing, MissingException, levels, skipmissing
+export allowmissing, disallowmissing, ismissing, missing, missings,
+       Missing, MissingException, levels, skipmissing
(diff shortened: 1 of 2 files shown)


src/Missings.jl
```diff
@@ -50,6 +51,31 @@ missings(dims...) = fill(missing, dims)
...
+    allowmissing(x::AbstractArray)
+
+Return an array equal to `x` allowing for [`missing`](@ref) values,
+i.e. with an element type equal to `Union{eltype(x), Missing}`.
+
+When possible, the result will share memory with `x` (as with [`convert`](@ref)).
+
+See also: [`disallowmissing`](@ref)
+"""
+allowmissing(x::AbstractVector{T}) where {T} = convert(AbstractArray{Union{T, Missing}}, x)
```
cjprybol: Oops. Didn't catch this until after I merged. I think these are supposed to be `x::AbstractArray` as shown in the docstring. I'll open a PR with a fix

PR #67, opened 2017-12-05
Fix (dis)allowmissing argument type (AbstractVector -> AbstractArray)

cjprybol:
My bad on missing this in the review of https://github.com/JuliaData/Missings.jl/pull/64


nalimilan:
Good catch!

PR #68, opened 2017-12-08
Use Base.Missing on Julia 0.7

nalimilan:
Adapt to https://github.com/JuliaLang/julia/pull/24653. Tests still fail on 0.7 due to a `promote_type` bug in Julia, but most features work.

### DIFF
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -1,50 +1,215 @@
 __precompile__(true)
 module Missings
 
-import Base: *, <, ==, !=, <=, !, +, -, ^, /, &, |, xor
 using Compat
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
        Missing, MissingException, levels, skipmissing
 
-"""
-    Missing
+if VERSION < v"0.7.0-DEV.2762"        
+    """
+        Missing
 
-A type with no fields whose singleton instance [`missing`](@ref) is used
-to represent missing values.
-"""
-struct Missing end
+    A type with no fields whose singleton instance [`missing`](@ref) is used
+    to represent missing values.
+    """
+    struct Missing end
 
-"""
-    missing
+    """
+        MissingException(msg)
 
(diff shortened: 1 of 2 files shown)


src/Missings.jl
```diff
@@ -1,51 +1,216 @@
...
 
-import Base: *, <, ==, !=, <=, !, +, -, ^, /, &, |, xor
 using Compat
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
        Missing, MissingException, levels, skipmissing
 
-"""
-    Missing
+if VERSION < v"0.7.0-DEV.2762"
```
nalimilan: It's kind of weird, but if I define the types in the same `if` block as the rest I get ` type definition not allowed inside a local scope` errors.

ararslan: That's really bizarre. What if you put an `@static` annotation on the `if`?

nalimilan: Same. AFAIK `@static` has no effect at the top level. I've found this issue, but it's pretty old and apparently closed: https://github.com/JuliaLang/julia/issues/10555

nalimilan:
I guess so. However it's not clear whether we'll be able to move everything to Base. For example I'm not sure `missings`, `allowmissing` and `disallowmissing!` will be accepted. We'll see...

src/Missings.jl
omus: `@static` is used at the top level in Compat.jl: https://github.com/JuliaLang/Compat.jl/blob/master/src/Compat.jl

nalimilan: Yes, but I can swear that I've read a comment by Jameson saying that has no effect.

omus: I used it to hide old deprecated syntax from 0.7: https://github.com/JuliaLang/Compat.jl/pull/419/files

ararslan: My understanding is that since the condition is evaluated at parse time, it gives an opportunity to trick the parser into ignoring things doesn't particularly like.

PR #70, opened 2018-01-10
Some compat for 0.7

quinnj:


test/runtests.jl
```diff
@@ -17,7 +22,7 @@ using Base.Test, Missings, Compat
     @test promote_type(Union{Int, Missing}, Union{Int, Missing}) == Union{Int, Missing}
     @test promote_type(Union{Float64, Missing}, Union{String, Missing}) == Any
     @test promote_type(Union{Float64, Missing}, Union{Int, Missing}) == Union{Float64, Missing}
-    @test promote_type(Union{Void, Missing, Int}, Float64) == Any
+    # @test_broken promote_type(Union{Nothing, Missing, Int}, Float64) == Any
```
quinnj: @nalimilan, are we maybe missing a `promote_rule` in the Base definitions? I'm seeing
```julia
julia> promote_type(Union{Nothing, Missing, Int}, Float64)
Union{Missing, Float64}
```
on 0.7 (as of today).

src/Missings.jl
```diff
@@ -206,15 +206,19 @@ else
     Base.float(A::AbstractArray{Missing}) = A
 end
 
+if isdefined(Base, :adjoint)
+Base.adjoint(d::Missing) = missing
```
ararslan: This should be defined in Base as well, right?

test/runtests.jl
```diff
@@ -1,4 +1,9 @@
-using Base.Test, Missings, Compat
+@static if VERSION < v"0.7.0-DEV.2005"
```
ararslan: This can just be `using Compat, Compat.Test, Missings`

src/Missings.jl
```diff
@@ -272,10 +276,17 @@ struct EachReplaceMissing{T, U}
     x::T
     replacement::U
 end
+if VERSION < v"0.7.0-DEV.3309"
 Base.iteratorsize(::Type{<:EachReplaceMissing{T}}) where {T} =
     Base.iteratorsize(T)
 Base.iteratoreltype(::Type{<:EachReplaceMissing{T}}) where {T} =
     Base.iteratoreltype(T)
+else
+Base.IteratorSize(::Type{<:EachReplaceMissing{T}}) where {T} =
```
ararslan: We should add this replacement to Compat.

quinnj: I was going to, but it's weird because `IteratorSize` is defined in 0.6 too.

src/Missings.jl
quinnj: Yeah, probably.

src/Missings.jl
ararslan: Oh it is? Hm, that is weird.

src/Missings.jl
quinnj: https://github.com/JuliaLang/julia/pull/25502

test/runtests.jl
nalimilan: That's https://github.com/JuliaLang/julia/pull/24653#discussion_r151874283. See also https://github.com/JuliaLang/julia/issues/25240.

src/Missings.jl
nalimilan: Also add indentation?

test/runtests.jl
nalimilan: But why comment this out? Isn't it still broken?

src/Missings.jl
```diff
@@ -286,6 +297,7 @@ Base.eltype(itr::EachReplaceMissing) = Missings.T(eltype(itr.x))
     (v isa Missing ? itr.replacement : v, s)
 end
 
+@static if !isdefined(Base, :skipmissing)
```
nalimilan: Why `@static` here and not in other places? I'm not sure whether it's needed.

ararslan: It's not necessary at the top level (though I think you can use it to get around deprecations from the parser) but it doesn't hurt.

test/runtests.jl
quinnj: It's not broken on 0.6 though

src/Missings.jl
ararslan: Fredrik is adding it in Compat as `Compat.IteratorSize`, so we could use that if this PR is merged after his is merged and tagged.

test/runtests.jl
nalimilan: How did tests pass then? :-/

nalimilan: OK. Why not make it conditional on the version then?

quinnj:
pushed changes to #72 instead.

PR #72, opened 2018-01-18
Export coalesce

### LINKED ISSUE #71: coalesce isn't exported
I'm not sure if there was a conscious decision behind this or not, but `coalesce` is not exported. I had the hardest time figuring out why we didn't have this functionality in Julia already until I was told that the function most definitely lives in this package. Doing a REPL help search for `Missings.coalesce` indeed shows that it does. I thought I'd just bring this to your attention if it hadn't already been mentioned.

ararslan:
Fixes #71. Not sure why this wasn't exported before.

### DIFF
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -3,10 +3,10 @@ module Missings
 
 using Compat
 
-export allowmissing, disallowmissing, ismissing, missing, missings,
+export allowmissing, coalesce, disallowmissing, ismissing, missing, missings,
        Missing, MissingException, levels, skipmissing
 
-if VERSION < v"0.7.0-DEV.2762"        
+if VERSION < v"0.7.0-DEV.2762"
     """
         Missing
 
@@ -30,7 +30,7 @@ end
 import Base: ==, !=, <, *, <=, !, +, -, ^, /, &, |, xor
 
 if VERSION >= v"0.7.0-DEV.2762"
-    using Base: ismissing, missing, Missing, MissingException
+    using Base: coalesce, ismissing, missing, Missing, MissingException
 else
     """
         missing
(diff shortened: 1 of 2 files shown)


src/Missings.jl
```diff
@@ -30,7 +30,7 @@ end
 import Base: ==, !=, <, *, <=, !, +, -, ^, /, &, |, xor
 
 if VERSION >= v"0.7.0-DEV.2762"
-    using Base: ismissing, missing, Missing, MissingException
+    using Base: coalesce, ismissing, missing, Missing, MissingException
```
nalimilan: IIRC `coalesce` has been added later, when I removed `Nullable`, so it should be under a different version check.

quinnj:
I had a fix for this locally along w/ some other 0.7 compat. Any objections to merging?

src/Missings.jl
```diff
@@ -441,29 +471,7 @@ julia> coalesce.([missing, 1, missing], [0, 10, 5])
...
-    levs = convert(AbstractArray{T}, filter!(!ismissing, unique(x)))
-    if method_exists(isless, Tuple{T, T})
-        try; sort!(levs); end
-    end
-    levs
-end
-
-# Deprecations
-@deprecate skip(itr) skipmissing(itr) false
+Compat.coalesce(x::Missing) = missing
```
nalimilan: This should only be defined before Base includes `coalesce`, or it's going to overwrite the equivalent Base method.

quinnj: That assumes that Compat already defines `coalesce(x::Missing)`, which it currently doesn't (it's commented out). It really boils down to whether the actual compat definition is in Missings or Compat.

nalimilan: Right, but I was speaking about a different issue, which is that recent Julia 0.7 versions define that method, and that `Compat.coalesce = Base.coalesce` there.

quinnj: I don't think so? I don't see any overwrite happening on latest master?

ararslan: Easy enough to just wrap this in `is !isdefined(Base, :coalesce)` regardless.

PR #76, opened 2018-01-23
Allow use of `missings` with a dims Tuple

omus:
Behaviour before this PR:
```julia
julia> zeros((2, 3))
2×3 Array{Float64,2}:
 0.0  0.0  0.0
 0.0  0.0  0.0

julia> missings((2, 3))
ERROR: MethodError: no method matching fill(::Missings.Missing, ::Tuple{Tuple{Int64,Int64}})
...
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -263,8 +263,12 @@ using Compat.Test, Compat.SparseArrays, Missings, Compat
     @test isequal(missings(Union{Int, Missing}, 1, 2), [missing missing])
     @test missings(Union{Int, Missing}, 1, 2) isa Matrix{Union{Int, Missing}}
     @test Union{Int, Missing}[1,2,3] == (Union{Int, Missing})[1,2,3]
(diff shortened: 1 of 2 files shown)


omus:
Potentially we may also want to define:
```julia
missings(a::AbstractArray) = missings(eltype(a), size(a))
```
I can save that for another PR.

test/runtests.jl
```diff
@@ -263,6 +263,8 @@ using Compat.Test, Compat.SparseArrays, Missings, Compat
     @test isequal(missings(Union{Int, Missing}, 1, 2), [missing missing])
     @test missings(Union{Int, Missing}, 1, 2) isa Matrix{Union{Int, Missing}}
     @test Union{Int, Missing}[1,2,3] == (Union{Int, Missing})[1,2,3]
+    @test isequal(missings(Int, (1, 2)), [missing missing])
```
nalimilan: Would be good to also check the type (see above).

omus:
Bump

PR #81, opened 2018-03-08
0.7 compat

quinnj:


src/Missings.jl
```diff
@@ -19,7 +19,7 @@ if VERSION < v"0.7.0-DEV.2762"
         MissingException(msg)
 
     Exception thrown when a [`missing`](@ref) value is encountered in a situation
-    where it is not supported. The error message, in the `msg` field
+    where it is not ƒpported. The error message, in the `msg` field
```
quinnj: what the.....?

ararslan: As much as I love that, I pushed a commit to reinstate the actual word.

Issue #80, opened 2018-02-20
Missings.T is unusable when T1 is a UnionAll


iamed2:
```julia
julia [LibPQ]> u = Union{Array{Float64}, Missing}
Union{Array{Float64,N} where N, Missings.Missing}

julia [LibPQ]> Missings.T(u)
Array{Float64,N}

julia [LibPQ]> [21.0]::(Missings.T(u))
ERROR: TypeError: typeassert: expected Array{Float64,N}, got Array{Float64,1}

julia [LibPQ]> [21.0] isa Missings.T(u)
false
```

I think this behaviour would be ideal:

```julia
julia [LibPQ]> u = Union{Array{Float64}, Missing}
Union{Array{Float64,N} where N, Missings.Missing}

julia [LibPQ]> Missings.T(u)
Array{Float64,N} where N

julia [LibPQ]> [21.0]::(Missings.T(u))
1-element Array{Float64,1}:
 21.0

julia [LibPQ]> [21.0] isa Missings.T(u)
true
```

nalimilan:
On 0.7, the behavior is different (maybe worse, but at least less weird). I think it has to do with dispatch:
```julia
julia> T(::Type{Union{T1, Missing}}) where {T1} = T1
T (generic function with 1 method)

julia> u = Union{Array{Float64}, Missing}
Union{Missing, Array{Float64,N} where N}

julia> T(u)
ERROR: MethodError: no method matching T(::Type{Union{Missing, Array{Float64,N} where N}})
Closest candidates are:
  T(::Type{Union{Missing, T1}}) where T1 at REPL[2]:1
```

Maybe file an issue in Julia?

Maybe we could work around this problem with a method dispatching on `::Union`. Could be worth investigating.

PR #82, opened 2018-03-17
Use typesubtract for Missings.T()

nalimilan:
It works in all cases where the old method works, but also fixes more cases, and works around a Julia bug which completely breaks the current definition in Julia 0.7.

Fixes https://github.com/JuliaData/Missings.jl/issues/80.

Cc: @iamed2


nalimilan:
OK to merge?

PR #89, opened 2018-08-30
Add conditional and passmissing

bkamins:
This is a proposal of implementation of https://github.com/JuliaLang/julia/pull/26661. Actually the best name for `conditional` would be `ifelse` but unfortunately it cannot be extended.

### DIFF
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -1,8 +1,7 @@
 module Missings
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
-       Missing, MissingException, levels, coalesce,
-       conditional, passmissing
+       Missing, MissingException, levels, coalesce, passmissing
 
 using Base: ismissing, missing, Missing, MissingException
 
@@ -166,32 +165,10 @@ function levels(x)
     levs
 end
 
-struct Conditional{P,X,Y} <: Function end
+struct PassMissing{F} <: Function end
 
-"""
-    conditional(predicate, x, y)
-
-Return a function that that applies function `predicate` to its positional and keyword
-arguments and returns the value of `x` applied to those arguments if `predicate`
-returns true and otherwise returns the value of `y` applied to those arguments.
-
-# Examples
-```jldoctest
-julia> f = conditional(x -> x ≥ 0, sqrt, x -> sqrt(complex(x)));
-
-julia> f.([4, -4])
-2-element Array{Number,1}:
-     2.0
-      0.0 + 2.0im
-"""
-conditional(predicate::Function, x::Base.Callable, y::Base.Callable) =
-    Conditional{predicate, x, y}()
-(::Conditional{P,X,Y})(xs...;kw...) where {P,X,Y} =
-    P(xs...; kw...) ? X(xs...; kw...) : Y(xs...; kw...)
-
-_passmissing_predicate(xs...;kw...) =
-    any(ismissing.(xs)) || any(ismissing.(values(values(kw))))
-_passmissing_value(xs...;kw...) = missing
+(::PassMissing{F})(xs...;kw...) where {F} =
+     any(ismissing, xs) || any(ismissing, values(values(kw))) ? missing : F(xs...; kw...)
 
 """
     passmissing(f)
@@ -206,6 +183,6 @@ julia> passmissing(sqrt).([missing, 4])
   missing
    2.0
 """
-passmissing(f) = conditional(_passmissing_predicate, _passmissing_value, f)
+passmissing(f::Base.Callable) = PassMissing{f}()
 
 end # module
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -165,16 +165,35 @@ function levels(x)
     levs
 end
 
-struct PassMissing{F} <: Function end
+struct PassMissing{F} <: Function
+    f::F
+end
 
-@generated (::PassMissing{F})(xs...;kw...) where {F} =
-    :(any(ismissing, xs) || any(ismissing, values(values(kw))) ? missing : F(xs...; kw...))
+function (f::PassMissing{F})(x) where {F}
+    if @generated
+        return x === Missing ? missing : :(f.f(x))
+    else
+        return x === missing ? missing : f.f(x)
+    end
+end
+
+function (f::PassMissing{F})(xs...) where {F}
+    if @generated
+        for T in xs
+            T === Missing && return missing
+        end
+        return :(f.f(xs...))
+    else
+        return any(ismissing, xs) ? missing : f.f(xs...)
+    end
+end
 
 """
     passmissing(f)
 
-Return a function that returns `missing` if any of its positional or keyword arguments
-are `missing` and otherwise applies `f` to those arguments.
+Return a function that returns `missing` if any of its positional arguments
+are `missing` (even if their number or type is not consistent with any of the
+methods defined for `f`) and otherwise applies `f` to these arguments.
 
 # Examples
 ```jldoctest
@@ -188,7 +207,13 @@ julia> passmissing(sqrt).([missing, 4])
 2-element Array{Union{Missing, Float64},1}:
   missing
    2.0
+
+julia> passmissing((x,y)->"$x $y")(1, 2)
+"1 2"
+
+julia> passmissing((x,y)->"$x $y")(missing)
+missing
 """
-passmissing(f::Base.Callable) = PassMissing{f}()
+passmissing(f::Base.Callable) = PassMissing{typeof(f)}(f)
 
 end # module
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -167,8 +167,8 @@ end
 
 struct PassMissing{F} <: Function end
 
-(::PassMissing{F})(xs...;kw...) where {F} =
-     any(ismissing, xs) || any(ismissing, values(values(kw))) ? missing : F(xs...; kw...)
+@generated (::PassMissing{F})(xs...;kw...) where {F} =
+    :(any(ismissing, xs) || any(ismissing, values(values(kw))) ? missing : F(xs...; kw...))
 
 """
     passmissing(f)
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -1,7 +1,8 @@
 module Missings
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
-       Missing, MissingException, levels, coalesce
+       Missing, MissingException, levels, coalesce,
+       conditional, passmissing
 
 using Base: ismissing, missing, Missing, MissingException
 
@@ -164,5 +165,47 @@ function levels(x)
     end
     levs
 end
+
+struct Conditional{P,X,Y} <: Function end
+
+"""
+    conditional(predicate, x, y)
+
+Return a function that that applies function `predicate` to its positional and keyword
+arguments and returns the value of `x` applied to those arguments if `predicate`
+returns true and otherwise returns the value of `y` applied to those arguments.
+
+# Examples
+```jldoctest
+julia> f = conditional(x -> x ≥ 0, sqrt, x -> sqrt(complex(x)));
+
+julia> f.([4, -4])
+2-element Array{Number,1}:
+     2.0
+      0.0 + 2.0im
+"""
+conditional(predicate::Function, x::Base.Callable, y::Base.Callable) =
+    Conditional{predicate, x, y}()
+(::Conditional{P,X,Y})(xs...;kw...) where {P,X,Y} =
+    P(xs...; kw...) ? X(xs...; kw...) : Y(xs...; kw...)
+
+_passmissing_predicate(xs...;kw...) =
+    any(ismissing.(xs)) || any(ismissing.(values(values(kw))))
+_passmissing_value(xs...;kw...) = missing
+
+"""
+    passmissing(f)
+
+Return a function that returns `missing` if any of its positional or keyword arguments
+are `missing` and otherwise applies `f` to those arguments.
+
+# Examples
+```jldoctest
+julia> passmissing(sqrt).([missing, 4])
+2-element Array{Union{Missing, Float64},1}:
+  missing
+   2.0
+"""
+passmissing(f) = conditional(_passmissing_predicate, _passmissing_value, f)
 
 end # module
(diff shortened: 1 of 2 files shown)


src/Missings.jl
```diff
@@ -165,4 +166,46 @@ function levels(x)
     levs
 end
 
+struct Conditional{P,X,Y} <: Function end
+
+"""
+    conditional(predicate, x, y)
```
nalimilan: What's the interest of having such a generic function here? Wouldn't it be simpler to hardcode the behavior of `passmissing`?

bkamins: The first reason is that I understand that this is "testing phase" of the concept. The function could be useful in Base for other reasons in the future, e.g. to provide similar functionality for `nothing`.

The second reason is that, as I have indicated in the main thread https://github.com/JuliaLang/julia/pull/26661, even for `missing` a user might want a specific behavior. For instance:

`passmissing(string)(10, pad=missing)`

would produce `missing`, but someone might want `missing` in `pad`  to be treated in a different way (e.g. throw an error) and write something like

```
condidional((x;kw...) -> ismissing(x), (x;kw...) -> missing, string)
```

But I will not stick with it if people do not find it useful.

nalimilan: I see your point, but Missings.jl only deals with missing values, so `nothing` is out of scope here. And given that the amount of code is small, even duplication wouldn't be too bad (actually, the docstring for `conditional` takes most of the lines).

bkamins: OK - so I understand we implement only a specialized `passmissing` for now? If yes I will amend the PR.

nalimilan: I'd say so, but you can wait for more input from others.

src/Missings.jl
```diff
@@ -165,4 +166,46 @@ function levels(x)
...
+     2.0
+      0.0 + 2.0im
+"""
+conditional(predicate::Function, x::Base.Callable, y::Base.Callable) =
+    Conditional{predicate, x, y}()
+(::Conditional{P,X,Y})(xs...;kw...) where {P,X,Y} =
+    P(xs...; kw...) ? X(xs...; kw...) : Y(xs...; kw...)
+
+_passmissing_predicate(xs...;kw...) =
+    any(ismissing.(xs)) || any(ismissing.(values(values(kw))))
```
ararslan: It's more efficient to pass `ismissing` as a function argument to `any` rather than broadcasting, i.e.
```julia
    any(ismissing, xs) || any(ismissing, values(values(kw)))
```

nalimilan: Right. In any case, this really needs to be unrolled, specializing on the number of arguments, or it will be too slow.

src/Missings.jl
```diff
@@ -165,4 +165,24 @@ function levels(x)
...
+are `missing` and otherwise applies `f` to those arguments.
+
+# Examples
+```jldoctest
+julia> passmissing(sqrt).([missing, 4])
+2-element Array{Union{Missing, Float64},1}:
+  missing
+   2.0
+"""
+passmissing(f::Base.Callable) = PassMissing{f}()
```
nalimilan: Shouldn't this be `typeof(f)`?

BTW, have you done some benchmarking to check there are no allocations?

src/Missings.jl
```diff
@@ -165,4 +165,24 @@ function levels(x)
...
+
+"""
+    passmissing(f)
+
+Return a function that returns `missing` if any of its positional or keyword arguments
+are `missing` and otherwise applies `f` to those arguments.
+
+# Examples
+```jldoctest
+julia> passmissing(sqrt).([missing, 4])
```
nalimilan: It would be useful to start with a simpler example with a single scalar (first non-missing, and then missing).

bkamins: added

src/Missings.jl
bkamins: AFAIK if I wanted to use `typeof{f}` then I would have to store `f` in a struct which would be slower and would allocate more.

I will post benchmarks for the current implementation in the main thread so that they do not disappear.

bkamins:
Here are the benchmarks. In general, if I understand it correctly, splatting slows it down in more complex scenarios. Fortunately for single positional argument this is fast (and I would say this is a most common case). We could define functions like:
```
(::PassMissing{F})(x1, x2) where {F} =
    ismissing(x1) || ismissing(x2) ? missing : F(x1, x2)
```
to handle more complex cases faster. I do not know if we want to as, hopefully, Julia should be able to handle such splatting as used in this PR fast (not sure though).

### Benchmark 1: simple scenario
```
julia> checkedsqrt(x) = x isa Missing ? missing : sqrt(x)
checkedsqrt (generic function with 1 method)

julia> @btime checkedsqrt(4.0)
  1.866 ns (0 allocations: 0 bytes)
2.0

julia> @btime checkedsqrt(missing)
  0.001 ns (0 allocations: 0 bytes)
missing

julia> checkedsqrt2(x) = passmissing(sqrt)(x)
checkedsqrt2 (generic function with 1 method)

julia> @btime checkedsqrt2(4.0)
  1.866 ns (0 allocations: 0 bytes)
2.0

julia> @btime checkedsqrt2(missing)
  0.001 ns (0 allocations: 0 bytes)
missing
```

### Benchmark 2: complex scenario
```
julia> parsemissing(t, s, b) = t isa Missing || s isa Missing || b isa Missing ? missing : parse(t, s, base=b)
parsemissing (generic function with 1 method)

julia> @btime parsemissing(Int, "1234", 5)
  89.091 ns (0 allocations: 0 bytes)
194

julia> @btime parsemissing(Int, "1234", missing)
  0.001 ns (0 allocations: 0 bytes)
missing

julia> @btime parsemissing(missing, "1234", 5)
  0.001 ns (0 allocations: 0 bytes)
missing

julia> parsemissing2(t, s, b) = passmissing(parse)(t, s, base=b)
parsemissing2 (generic function with 1 method)

julia> @btime parsemissing2(Int, "1234", 5)
  752.883 ns (4 allocations: 96 bytes)
194

julia> @btime parsemissing(Int, "1234", missing)
  0.001 ns (0 allocations: 0 bytes)
missing

julia> @btime parsemissing(missing, "1234", 5)
  0.001 ns (0 allocations: 0 bytes)
missing
```

bkamins:
Using `@generated` improves performance a bit.

bkamins:
Of course some independent benchmarks are welcome (but at least looking at `@code_warntype` the generated version is type-inferred correctly as should be expected).

CC @pszufe

src/Missings.jl
```diff
@@ -1,7 +1,7 @@
 module Missings
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
-       Missing, MissingException, levels, coalesce
+       Missing, MissingException, levels, coalesce, passmissing
```
quinnj: I personally prefer `Missings.propagate` if we're still bikeshedding the name here.

src/Missings.jl
```diff
@@ -165,4 +165,30 @@ function levels(x)
     levs
 end
 
+struct PassMissing{F} <: Function end
+
+@generated (::PassMissing{F})(xs...;kw...) where {F} =
```
quinnj: I don't think the `@generated` is doing anything here except ensuring a new function gets compiled w/ every call.

src/Missings.jl
```diff
@@ -165,4 +165,30 @@ function levels(x)
     levs
 end
 
+struct PassMissing{F} <: Function end
```
quinnj: I think we probably want:
```julia
struct PassMissing{F} <: Function
    f::F
end
```

src/Missings.jl
quinnj: I think in most practical uses, the struct would never allocate. It's the same idea w/ the new iteration protocol: every call to `iterate` potentially returns a `Tuple{T...}`, but most of the time those tuples and even intermediate allocated objects don't even get allocated.

src/Missings.jl
```diff
@@ -165,4 +165,30 @@ function levels(x)
...
+
+@generated (::PassMissing{F})(xs...;kw...) where {F} =
+    :(any(ismissing, xs) || any(ismissing, values(values(kw))) ? missing : F(xs...; kw...))
```
quinnj: Hmmm, the inclusion of keyword arguments here seems off to me; what if `missing` is a totally valid thing to pass as a keyword argument, but now it's interacting weird w/ `PassMissing` because it always makes my result `missing`?

quinnj: I think my stab at this definition would be more along the lines of:
```julia
struct PassMissing{F} <: Function                                                                                             
    f::F                                                                                                                      
end                                                                                                                           
                                                                                                                              
function (f::PassMissing{F})(x) where {F}                                                                                     
    if @generated                                                                                                             
        return x === Missing ? missing : :(f.f(x))                                                                            
    else                                                                                                                      
        return x === missing ? missing : f.f(x)                                                                               
    end                                                                                                                       
end                                                                                                                           
                                                                                                                              
function (f::PassMissing{F})(xs...; kw...) where {F}                                                                          
    if @generated                                                                                                             
        for T in xs                                                                                                           
            T === Missing && return missing                                                                                   
        end                                                                                                                   
        return :(f.f(xs...; kw...))                                                                                           
    else                                                                                                                      
        return any(ismissing, xs) ? missing : f.f(xs...; kw...)                                                               
    end                                                                                                                       
end 
```
with specialize methods for 1 argument, 2 argument, maybe 3.

nalimilan: Interesting. Though with generated functions you don't need to specialize on the number of arguments.

Also why do you think keyword arguments should be excluded from the missingness check?

quinnj: The `if @generated` part is just a future-proof against a time when you're able to statically compile a julia program into an executable that can run without LLVM (i.e. w/o runtime compilation). The `else` block just provides a path that would be executed if the compiler encountered an already non-generated/compiled method for the given arguments at runtime.

For keyword arguments, it just feels more arbitrary. Like, it's hard to imagine a case where I'd be `map`ing over some `Vector{Union{Float64, Missing}}` and `Vector{Union{Int, Missing}}` and be doing something like `map((x, y)->round(x; digits=y), zip(A, B))`. i.e. when would I maybe pass a value vs. maybe pass a `missing` as a keyword argument? Keyword arguments also tend to use various "sentinel" values as signals or special values for the function to use, including `missing`; the danger there being that someone writes a function `foo(x...; sentinel=nothing)`, but when the user tries to call `Missings.propagate(foo)(x; sentinel=missing)`, the entire result comes back `missing` instead of passing `missing` on to `foo` as a valid sentinel value.

If anything, I'd say we leave keyword arguments out of the mix for now since they can always be added later.

My main issue with `passmissing` is that `pass` usually has a connotation of "ignore" or "do nothing" (see python's [pass](https://docs.python.org/2.0/ref/pass.html)), whereas we're not really ignoring `missing` values, we're propagating them.

nalimilan: > For keyword arguments, it just feels more arbitrary. Like, it's hard to imagine a case where I'd be `map`ing over some `Vector{Union{Float64, Missing}}` and `Vector{Union{Int, Missing}}` and be doing something like `map((x, y)->round(x; digits=y), zip(A, B))`. i.e. when would I maybe pass a value vs. maybe pass a `missing` as a keyword argument? Keyword arguments also tend to use various "sentinel" values as signals or special values for the function to use, including `missing`; the danger there being that someone writes a function `foo(x...; sentinel=nothing)`, but when the user tries to call `Missings.propagate(foo)(x; sentinel=missing)`, the entire result comes back `missing` instead of passing `missing` on to `foo` as a valid sentinel value.

I don't think there are any examples of functions using `missing` as a sentinel currently, right? We rather want people to use `nothing` for that.

I see your point about keyword arguments generally being about options, while the data is passed as positional arguments, but I'd find it problematic to make this a rule. Overall it doesn't seem we have strong use cases for either behavior, and in such situations I tend to favor the simplest rule (i.e. "return `missing` if one of the arguments is `missing`").

> If anything, I'd say we leave keyword arguments out of the mix for now since they can always be added later.

The problem is, changing that would be breaking. The only way to be able to choose any behavior later is to throw an error if a keyword argument is `missing`.

src/Missings.jl
nalimilan: I like `propagate` too, the only gripe is that it's quite long. But well...

src/Missings.jl
bkamins: > The only way to be able to choose any behavior later is to throw an error if a keyword argument is missing.

We could define lift for functions that do not take keyword arguments for the time being. It is easy enough to wrap a function requiring kwargs in an anonymous function.

src/Missings.jl
bkamins: But do we want to export it or not? If yes then `propagate` is ok, but `Missings.propagate` is a bit long.

nalimilan: Unfortunately `propagate` is quite general to claim it for this feature...

bkamins: I know, but that is why I thought the idea was `propagatemissing` which tab-completes.

tcovert:
Did this ever make its way into master?  Having a consistent way to "wrap" routines that might have to deal with missing values in DataFrames data cleaning would be pretty nice.  Right now, the absence of missing-aware Date methods is kind of a pain...

bkamins:
I was not clear what the best API would be so I left it hanging.

@tcovert do you have any opinion what would be best given the discussion above?

bkamins:
Having looked at the options I recommend:
1. to use the design proposed by @quinnj
2. not to accept keyword arguments at all; we can add this support later (that change will be non-breaking) and there is no strong consensus what to do with them for now; I would say that most use cases will not involve keyword arguments anyway

@nalimilan I we accept this PR the side effect will be that DataFrames.jl will re-export this (which is actually nice I think) - I will add it to the documentation and tutorial.

bkamins:
codecov gave a strange result (and the `else` part of `if @generated` is not covered by tests)

nalimilan:
Sounds good for now. Can you document (and possibly test) that keyword arguments are not supported?

bkamins:
Added. Thanks for looking at it.

bkamins:
Do we want to merge it or wait for more feedback?

nalimilan:
There's been plenty of time for feedback, and the PR throws an error for keyword arguments, so we can discuss that later.

PR #98, opened 2019-06-24
Update README for Julia 1.0

### LINKED ISSUE #97: skipping missing values
Hello,

according doc

> In many cases, missing values will have to be skipped or replaced with a valid value. For example, sum([1, missing]) returns missing due to the behavior of +. Use sum(Missings.skip([1, missing]) to ignore missing values. 

    sum(Missings.skip([1, missing])

is problematic (ending parenthesis is missing) but

    sum(Missings.skip([1, missing]))

raises

    ERROR: MethodError: no method matching skip(::Array{Union{Missing, Int64},1})

Doc (and maybe code also) should be fixed.

Kind regards

nalimilan:
Fixes #97.

We could also add deprecations for remaining functions, but probably better do that for all functions at the same time once replacements have been included in Julia (notably `nonmissingtype` and `passmissing`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,9 @@
 
 # Missings
 
-*A missing value representation for Julia for databases and statistics*
+*Convenience functions to work with missing values in Julia*
 
-| **PackageEvaluator**                                            | **Build Status**                                                                                |
-|:---------------------------------------------------------------:|:-----------------------------------------------------------------------------------------------:|
-|[![][pkg-0.6-img]][pkg-0.6-url] | [![][travis-img]][travis-url] [![][appveyor-img]][appveyor-url] [![][codecov-img]][codecov-url] |
+[![][travis-img]][travis-url] [![][appveyor-img]][appveyor-url] [![][codecov-img]][codecov-url]
 
 
 ## Installation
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,10 +16,15 @@ julia> Pkg.add("Missings")
 
 ## Project Status
 
-The package is tested against the current Julia `0.6` release and nightly on Linux, OS X, and Windows.
-
 Starting from Julia `0.7` the `Missing` type and basic functionalities related to it have been merged to core Julia.
-This package still provides additional features to handle `missing` values.
+See [the Julia manual](https://docs.julialang.org/en/v1/manual/missing/index.html) for general documentation on missing values in Julia.
+This package still provides additional features to handle `missing` values:
+- `Missings.T` to extract `T` from a `Union{T, Missing}` type
```
ararslan: Is it worth mentioning this at all given that `Base.nonmissingtype` exists?

nalimilan: Until it's exported I'd rather not recommend it. Also it's broken in complex cases in older Julia versions.

ararslan: That's fair I suppose.

nickrobinson251: #102

nickrobinson251:
```suggestion
- `Missings.nonmissingtype` to extract `T` from a `Union{T, Missing}` type
```

PR #99, opened 2019-08-28
Don't use Base.Callable

nickrobinson251:
it's not exported. i think we can survive without it.

### DIFF
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -223,6 +223,6 @@ julia> passmissing((x,y)->"\$x \$y")(1, 2)
 julia> passmissing((x,y)->"\$x \$y")(missing)
 missing
 """
-passmissing(f::Base.Callable) = PassMissing{typeof(f)}(f)
+passmissing(f) = PassMissing{typeof(f)}(f)
 
 end # module
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -223,6 +223,6 @@ julia> passmissing((x,y)->"\$x \$y")(1, 2)
 julia> passmissing((x,y)->"\$x \$y")(missing)
 missing
 """
-passmissing(f) = PassMissing{typeof(f)}(f)
+passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
 end # module
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -139,11 +139,16 @@ using Test, SparseArrays, Missings
     @test_throws MethodError disallowmissing([missing missing])
 
     # Lifting
+    ## functor
     struct CubeRooter end
     (::CubeRooter)(x) = cbrt(x)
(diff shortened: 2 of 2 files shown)


bkamins:
I am OK with this change.

rofinn:
@bkamins Should this be merged then?

src/Missings.jl
```diff
@@ -216,6 +216,6 @@ julia> passmissing((x,y)->"\$x \$y")(1, 2)
 julia> passmissing((x,y)->"\$x \$y")(missing)
 missing
 """
-passmissing(f::Base.Callable) = PassMissing{typeof(f)}(f)
+passmissing(f) = PassMissing{typeof(f)}(f)
```
ararslan: Unrelated to this PR, but I think `typeof(f)` here should actually be `Core.Typeof(f)`.
```julia
julia> typeof(Int)
DataType

julia> Core.Typeof(Int)
Type{Int64}

julia> typeof(sin) === Core.Typeof(sin)
true
```

nickrobinson251: will open another PR :)

nickrobinson251: (after this one is merged)

bkamins: Good point - I would propose to change this in this PR and add appropriate tests for: function, type and functor.

@nickrobinson251 - would you be willing to do this?

nickrobinson251: 👍

bkamins:
I propose to wait for @nalimilan to comment on this. After merging all the recent PRs probably we should bump a version of Missings.jl and tag a release.

src/Missings.jl
nickrobinson251: Let me now if I missed anything 😄

test/runtests.jl
```diff
@@ -139,13 +139,22 @@ using Test, SparseArrays, Missings
...
+    @test passmissing(cuberoot)(27) == 3.0
+    @test isequal(passmissing(cuberoot)(missing), missing)
     @test passmissing(sqrt)(4) == 2.0
     @test isequal(passmissing(sqrt)(missing), missing)
     @test isequal(passmissing(sqrt).([missing, 4]), [missing, 2.0])
     @test passmissing((x,y)->"$x $y")(1, 2) == "1 2"
     @test isequal(passmissing((x,y)->"$x $y")(missing), missing)
     @test_throws ErrorException passmissing(string)(missing, base=2)
 
+    @test passmissing(sin) === Missings.PassMissing{typeof(sin)}(sin)
```
bkamins: We also might add tests `passmissing(Int)(1)` and `passmissing(Int)(missing)`

@ararslan Do you think that with `Core.Typeof` the speed of the above will increase?

nickrobinson251: added :)

nalimilan:
Thanks!

PR #102, opened 2019-08-30
Deprecate `Missings.T` -> `Missings.nonmissingtype`

### LINKED ISSUE #101: `Missings.T` can use Base's `nonmissingtype` on Julia 1.3+
~There is nothing for us to do right now, but...~

As of Julia 1.3, [Base will export `nonmissingtype`](https://github.com/JuliaLang/julia/blob/768b25f6a850bf68e4c6c85ede37c44b16c79625/base/missing.jl#L21-L39)

Which ([since Julia 1.2](https://github.com/JuliaLang/julia/commit/98673b5a0f4efd72dd401b603745c70193e1f64b)) is the same as `Missings.T`
https://github.com/JuliaData/Missings.jl/blob/3ac53e982337b5c3e9ab625f69e3b5454b89d422/src/Missings.jl#L8

In future we should consider deprecating `Missing.T` and point users to `nonmissingtype`

nickrobinson251:
- closes #101 
- fits with Base's naming -- [nonmissingtype is exported from Base in Julia 1.3](https://github.com/JuliaLang/julia/commit/1e6d5f1a5b3041eba854412afa813b903182fa5b)
- Base.nonmissingtype exactly matches the definition of Missings.T ([since Julia 1.2](https://github.com/JuliaLang/julia/commit/98673b5a0f4efd72dd401b603745c70193e1f64b))
- Adds CI testing on 1.2 and 1.3
- Missings.T was not exported, so Missings.nonmissingtype is not exported
- Let me know if and how I should bump the version number

### DIFF
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -4,8 +4,15 @@ export allowmissing, disallowmissing, ismissing, missing, missings,
        Missing, MissingException, levels, coalesce, passmissing
 
 using Base: ismissing, missing, Missing, MissingException
+using Base: @deprecate
 
-T(::Type{S}) where {S} = Core.Compiler.typesubtract(S, Missing)
+@static if VERSION < v"1.3.0-alpha.121"
+    nonmissingtype(::Type{S}) where {S} = Core.Compiler.typesubtract(S, Missing)
+else
+    using Base: nonmissingtype
+end
+
+@deprecate T nonmissingtype false
 
 # vector constructors
 missings(dims::Dims) = fill(missing, dims)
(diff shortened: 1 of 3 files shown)


src/Missings.jl
```diff
@@ -4,8 +4,15 @@ export allowmissing, disallowmissing, ismissing, missing, missings,
...
+using Base: @deprecate
 
-T(::Type{S}) where {S} = Core.Compiler.typesubtract(S, Missing)
+@static if VERSION < v"1.3.0-alpha.121"
+    nonmissingtype(::Type{S}) where {S} = Core.Compiler.typesubtract(S, Missing)
+else
+    using Base: nonmissingtype
+end
+
+@deprecate T nonmissingtype false
```
rofinn: RIP :(

nalimilan:
We'd better also export `nonmissingtype`, since in Julia 1.3 the name will be available without the `Missings.` prefix. See https://github.com/JuliaData/Missings.jl/pull/105.

PR #105, opened 2019-09-04
Export nonmissingtype

nalimilan:
Cc: @nickrobinson251


nalimilan:
AFAIK reexporting a function imported from elsewhere has no effect.

nickrobinson251:
:+1: I'm in favour of this too, and was only trying to keep the first PR as simple as possible. Thanks for following up!

nalimilan:
I've updated JuliaRegistries/General/3224.

PR #108, opened 2019-10-03
Use levels from DataAPI

nalimilan:
Keep exporting it for now: if we added a deprecation, it would be triggered even when people call levels after loading CategoricalArrays or DataFrames.

Depends on JuliaData/DataAPI.jl#12.


nalimilan:
https://github.com/JuliaRegistries/General/pull/4057

PR #111, opened 2019-12-23
Add multiskipmissing

pdeffebach:
This PR adds functionality for `multiskipmissings` as discussed in Julia issue 30596 [here](https://github.com/JuliaLang/julia/issues/30596). 

The motivation for this PR was noticing that 

```
cor(x, y)
```

has no missing value support. 

```
julia> x = allowmissing(rand(10)); y = allowmissing(rand(10));

julia> x[1] = missing; y[2] = missing;

julia> cor(x, y)
missing
```

In order for this to work, you need to collect the non-missing matching values of `x` and `y`. To do that I implement the function `multiskipmissing`, which returns a `Tuple` of iterators. 

```
julia> x = [1, 2, missing, 4]; y = [1, 2, 3, missing];
julia> tx, ty = multiskipmissing(x, y)
julia> for t in tx
       @show t
       end
t = 1
t = 2
```

In the above example, each `tx` and `ty`  are `MultiSkipMissing` objects which have `zip(x, y)` as a field. I then implemented `skipmissing` from Base with some changed logic such that instead of checking if the array of interest has a missing element, I check if any elements of the `i`th element of `zip(x, y)` are missing. 

I this PR is not complete. I still can't get `collect` to work, so the original goal is not yet met.

### DIFF
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -207,4 +207,261 @@ missing
 """
 passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
+"""
+   skipmissings(args...)
+
+Return a tuple of iterators wrapping each of the iterators in `args`, but
+skipping elements at positions where at least one of the iterators returns `missing`
+(listwise deletion of missing values).
+
+# Examples
+```
+julia> x = [1, 2, missing, 4]; y = [1, 2, 3, missing];
+
+julia> tx, ty = skipmissings(x, y)
+(Missings.SkipMissings{Array{Union{Missing, Int64},1},Tuple{Array{Union{Missing, Int64},1}}}
+(Union{Missing, Int64}[1, 2, missing, 4], (Union{Missing, Int64}[1, 2, 3, missing],)), Missi
+ngs.SkipMissings{Array{Union{Missing, Int64},1},Tuple{Array{Union{Missing, Int64},1}}}(Union
+{Missing, Int64}[1, 2, 3, missing], (Union{Missing, Int64}[1, 2, missing, 4],)))
+
+julia> collect(tx)
+2-element Array{Int64,1}:
+ 1
+ 2
+
+```
+"""
+function skipmissings(args...)
+    if isempty(args)
+        throw(ArgumentError("Must input one or more arguments"))
+    end
+
+    if args isa Tuple{Vararg{AbstractArray}}
+        if !all(x -> length(x) == length(args[1]), args)
+            throw(ArgumentError("All arguments must have the same length"))
+        end
+
+        if !all(x -> eachindex(x) == eachindex(args[1]), args)
+            throw(ArgumentError("All arguments must have the same indices"))
+        end
+    end
+
+    ntuple(length(args)) do i
+        s = setdiff(1:length(args), i)
+        SkipMissings(args[i], args[s])
+    end
+end
+
+struct SkipMissings{V, T}
+    x::V
+    others::T
+end
+
+Base.@propagate_inbounds function _anymissingindex(others::Tuple{Vararg{AbstractArray}}, i)    
+   for oth in others
+        oth[i] === missing && return true
+    end
+
+    return false
+end
+
+@inline function _anymissingiterate(others::Tuple, state)
+    for oth in others 
+        y = iterate(oth, state)
+        y !== nothing && first(y) === missing && return true
+    end
+
+    return false
+end
+
+const SkipMissingsofArrays = SkipMissings{V, T} where
+    {V <: AbstractArray, T <: Tuple{Vararg{AbstractArray}}}
+
+function Base.show(io::IO, mime::MIME"text/plain", itr::SkipMissings{V}) where V 
+    print(io, SkipMissings, '{', V, '}', '(', itr.x, ')', " comprised of " *
+          "$(length(itr.others) + 1) iterators")
+end
+
+Base.IteratorSize(::Type{<:SkipMissings}) = Base.SizeUnknown()
+Base.IteratorEltype(::Type{<:SkipMissings{V}}) where {V} = Base.IteratorEltype(V)
+Base.eltype(::Type{<:SkipMissings{V}}) where {V} = nonmissingtype(eltype(V))
+Base.IndexStyle(itr::SkipMissings) = Base.IndexStyle(itr.x)
+
+function Base.iterate(itr::SkipMissings, state=1)
+    x_itr = iterate(itr.x, state)
+    x_itr === nothing && return nothing
+    x_item, x_state = x_itr
+    while true
+        x_item === missing || _anymissingiterate(itr.others, state) || break
+        x_itr = iterate(itr.x, x_state)
+        x_itr === nothing && return nothing
+        state = x_state
+        x_item, x_state = x_itr
+    end
+    return x_item, x_state
+end
+
+function Base.iterate(itr::SkipMissingsofArrays, state=0)
+    eix = eachindex(itr.x)
+    ind_itr = iterate(eix, state)
+    ind_itr === nothing && return nothing
+    ind_item, ind_state = ind_itr
+    @inbounds x_item = itr.x[ind_item]
+    @inbounds while true
+        x_item === missing || _anymissingindex(itr.others, ind_item) || break
+        ind_itr = iterate(eix, ind_state)
+        ind_itr === nothing && return nothing
+        ind_item, ind_state = ind_itr
+        x_item = itr.x[ind_item]
+    end
+    return x_item, ind_state
+end
+
+Base.IndexStyle(::Type{<:SkipMissings{V}}) where {V} = Base.IndexStyle(V)
+
+function Base.eachindex(itr::SkipMissingsofArrays)
+    @inbounds Iterators.filter(eachindex(itr.x)) do i
+        itr.x[i] !== missing && !_anymissingindex(itr.others, i)
+    end
+end
+
+function Base.keys(itr::SkipMissingsofArrays)
+    @inbounds Iterators.filter(keys(itr.x)) do i
+        itr.x[i] !== missing && !_anymissingindex(itr.others, i)
+    end
+end
+
+@inline function Base.getindex(itr::SkipMissingsofArrays, i)
+    @boundscheck checkbounds(itr.x, i)
+    @inbounds xi = itr.x[i]
+    if xi === missing || @inbounds _anymissingindex(itr.others, i) 
+        throw(MissingException("the value at index $i is missing for some element"))
+    end
+    return xi
+end
+
+Base.mapreduce(f, op, itr::SkipMissingsofArrays) =
+    Base._mapreduce(f, op, Base.IndexStyle(itr), itr)
+
+function Base._mapreduce(f, op, ::IndexLinear, itr::SkipMissingsofArrays)
+    A = itr.x
+    local ai
+    inds = LinearIndices(A)
+    i = first(inds)
+    ilast = last(inds)
+    @inbounds while i <= ilast
+        ai = A[i]
+        ai === missing || _anymissingindex(itr.others, i) || break
+        i += 1
+    end
+    i > ilast && return Base.mapreduce_empty(f, op, Base.eltype(itr))
+    a1 = ai
+    i += 1
+    @inbounds while i <= ilast
+        ai = A[i]
+        ai === missing || _anymissingindex(itr.others, i) || break
+        i += 1
+    end
+    i > ilast && return Base.mapreduce_first(f, op, a1)
+    # We know A contains at least two non-missing entries: the result cannot be nothing
+    something(Base.mapreduce_impl(f, op, itr, first(inds), last(inds)))
+end
+
+Base._mapreduce(f, op, ::IndexCartesian, itr::SkipMissingsofArrays) = mapfoldl(f, op, itr)
+
+
+Base.mapreduce_impl(f, op, A::SkipMissingsofArrays, ifirst::Integer, ilast::Integer) =
+    Base.mapreduce_impl(f, op, A, ifirst, ilast, Base.pairwise_blocksize(f, op))
+
+# Returns nothing when the input contains only missing values, and Some(x) otherwise
+@noinline function Base.mapreduce_impl(f, op, itr::SkipMissingsofArrays,
+                                       ifirst::Integer, ilast::Integer, blksize::Int)
+    A = itr.x
+    if ifirst == ilast
+        @inbounds a1 = A[ifirst]
+        if a1 === missing 
+            return nothing
+        elseif _anymissingindex(itr.others, ifirst) 
+            return nothing
+        else
+            return Some(Base.mapreduce_first(f, op, a1))
+        end
+    elseif ifirst + blksize > ilast
+        # sequential portion
+        local ai
+        i = ifirst
+        @inbounds while i <= ilast
+            ai = A[i]
+            ai === missing || _anymissingindex(itr.others, i) || break
+            i += 1
+        end
+        i > ilast && return nothing
+        a1 = ai::eltype(itr)
+        i += 1
+        @inbounds while i <= ilast
+            ai = A[i]
+            ai === missing || _anymissingindex(itr.others, i) || break
+            i += 1
+        end
+        i > ilast && return Some(Base.mapreduce_first(f, op, a1))
+        a2 = ai::eltype(itr)
+        i += 1
+        v = op(f(a1), f(a2))
+        @simd for i = i:ilast
+            @inbounds ai = A[i]
+            ai === missing || @inbounds _anymissingindex(itr.others, i) || (v = op(v, f(ai)))
+        end
+        return Some(v)
+    else
+        # pairwise portion
+        imid = (ifirst + ilast) >> 1
+        v1 = Base.mapreduce_impl(f, op, itr, ifirst, imid, blksize)
+        v2 = Base.mapreduce_impl(f, op, itr, imid+1, ilast, blksize)
+        if v1 === nothing && v2 === nothing
+            return nothing
+        elseif v1 === nothing
+            return v2
+        elseif v2 === nothing
+            return v1
+        else
+            return Some(op(something(v1), something(v2)))
+        end
+    end
+end
+
+"""
+    filter(f, itr::SkipMissings)
+
+Return a vector similar to the array wrapped by the given `SkipMissings` iterator
+but skipping all elements with a `missing` value in one of the iterators passed
+to `skipmissing` and elements for which `f` returns `false`. This method
+only applies when all iterators passed to `skipmissings` are arrays. 
+
+# Examples
+```
+julia> x = [missing; 2:9]; y = [1:9; missing];
+
+julia> mx, my = skipmissings(x, y);
+
+julia> filter(isodd, mx)
+4-element Array{Int64,1}:
+ 3
+ 5
+ 7
+ 9
+
+```
+"""
+function filter(f, itr::SkipMissingsofArrays)
+    x = itr.x
+    y = similar(x, eltype(itr), 0)
+    for i in eachindex(x)
+        @inbounds xi = x[i]
+        if xi !== missing && @inbounds !_anymissingindex(itr.others, i) && f(xi)
+            push!(y, xi)
+        end
+    end
+    y
+end
+
 end # module
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -1,7 +1,8 @@
 module Missings
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
-       Missing, MissingException, levels, coalesce, passmissing, nonmissingtype
+       Missing, MissingException, levels, coalesce, passmissing, nonmissingtype,
+       skipmissings
 
 using Base: ismissing, missing, Missing, MissingException
 using Base: @deprecate
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -123,11 +123,8 @@ struct CubeRooter end
     y = [isodd(i) ? missing : i for i in 65:128]
     mx, my = skipmissings(x, y)
     @test sum(mx) === 1056
-    if VERSION >= v"1.4.0-DEV"
-        t = quote
-            @inferred Union{Missing, Int} sum($mx)
-        end
-        eval(t)
+    @static if VERSION >= v"1.4.0-DEV"
+        @inferred Union{Missing, Int} sum(mx)
     end
 
     @test levels(1:1) == levels([1]) == levels([1, missing]) == levels([missing, 1]) == [1]
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -84,7 +84,7 @@ struct CubeRooter end
     iobuf = IOBuffer()
     show(iobuf, MIME("text/plain"), mx)
     s = String(take!(iobuf))
-    @test s == "Missings.SkipMissings{Array{Union{Missing, Int64},1}}(Union{Missing, Int64"*
+    @test s == "Missings.SkipMissings{Array{Union{Missing, Int$(Sys.WORD_SIZE)},1}}(Union{Missing, Int$(Sys.WORD_SIZE)" *
         "}[1, 2, missing, 4]) comprised of 2 iterators"
     @test collect(mx) == [1, 2]
     @test collect(mx) isa Vector{Int} 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -107,7 +107,7 @@ struct CubeRooter end
     @test isapprox(mapreduce(cos, *, collect(mx)),  mapreduce(cos, *, mx))
     if VERSION >= v"1.4.0-DEV"
         t = quote 
-            @inferred Union{Float64, Missing} mapreduce(cos, *, mx)
+            @inferred Union{Float64, Missing} mapreduce(cos, *, $mx)
         end
         eval(t)
     end
@@ -125,7 +125,7 @@ struct CubeRooter end
     @test sum(mx) === 1056
     if VERSION >= v"1.4.0-DEV"
         t = quote
-            @inferred Union{Missing, Int} sum(mx)
+            @inferred Union{Missing, Int} sum($mx)
         end
         eval(t)
     end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -106,7 +106,10 @@ struct CubeRooter end
     @test reduce(+, mx) === 18
     @test isapprox(mapreduce(cos, *, collect(mx)),  mapreduce(cos, *, mx))
     if VERSION >= v"1.4.0-DEV"
-        @inferred Union{Float64, Missing} mapreduce(cos, *, mx)
+        t = quote 
+            @inferred Union{Float64, Missing} mapreduce(cos, *, mx)
+        end
+        eval(t)
     end
 
     x = [missing missing missing]
@@ -121,7 +124,10 @@ struct CubeRooter end
     mx, my = skipmissings(x, y)
     @test sum(mx) === 1056
     if VERSION >= v"1.4.0-DEV"
-        @inferred Union{Missing, Int} sum(mx)
+        t = quote
+            @inferred Union{Missing, Int} sum(mx)
+        end
+        eval(t)
     end
 
     @test levels(1:1) == levels([1]) == levels([1, missing]) == levels([missing, 1]) == [1]
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -84,8 +84,8 @@ struct CubeRooter end
     iobuf = IOBuffer()
     show(iobuf, MIME("text/plain"), mx)
     s = String(take!(iobuf))
-    @test s == "Missings.SkipMissings{Array{Union{Missing, Int$(Sys.WORD_SIZE)},1}}(Union{Missing, Int$(Sys.WORD_SIZE)" *
-        "}[1, 2, missing, 4]) comprised of 2 iterators"
+    @test s == "Missings.SkipMissings{Array{Union{Missing, $Int},1}}(" *
+        "Union{Missing, $Int}[1, 2, missing, 4]) comprised of 2 iterators"
     @test collect(mx) == [1, 2]
     @test collect(mx) isa Vector{Int} 
     @test reduce(+, mx) === reduce(+, collect(mx)) === sum(mx) ===
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -120,7 +120,6 @@ struct CubeRooter end
     y = [isodd(i) ? missing : i for i in 65:128]
     mx, my = skipmissings(x, y)
     @test sum(mx) === 1056
-    @inferred Union{Missing, Int} sum(mx)
     if VERSION >= v"1.4.0-DEV"
         @inferred Union{Missing, Int} sum(mx)
     end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -75,6 +75,56 @@ struct CubeRooter end
     @test collect(x) == [1, 2, 4]
     @test collect(x) isa Vector{Int}
 
+    x = [1, 2, missing, 4]
+    y = ["a", "b", "c", missing]
+    z = [missing, missing, 3.1, 4.5]
+    l = [1, 2, 3, 4, 5]
+    @test_throws ArgumentError skipmissings(x, l)
+    mx, my = skipmissings(x, y)
+    iobuf = IOBuffer()
+    show(iobuf, MIME("text/plain"), mx)
+    s = String(take!(iobuf))
+    @test s == "Missings.SkipMissings{Array{Union{Missing, Int64},1}}(Union{Missing, Int64"*
+        "}[1, 2, missing, 4]) comprised of 2 iterators"
+    @test collect(mx) == [1, 2]
+    @test collect(mx) isa Vector{Int} 
+    @test reduce(+, mx) === reduce(+, collect(mx)) === sum(mx) ===
+        mapreduce(identity, +, mx) === 3
+    @test mapreduce(x -> x^2, +, mx) === mapreduce(x -> x^2, +, collect(mx)) === 5
+    mx, my, mz = skipmissings(x, y, z)
+    @test eltype(mx) == Int
+    @test eltype(my) == String
+    @test eltype(mz) == Float64
+    @test isempty(collect(mx))
+    @test sum(mx) === 0
+    x = [missing 4; 2 5; 3 6]
+    y = [1 4; missing 5; 3 6]
+    mx, my = skipmissings(x, y)
+    @test collect(mx) == [3, 4, 5, 6]
+    @test mx[3] == 3
+    @test_throws MissingException mx[1]
+    @test reduce(+, mx) === 18
+    @test isapprox(mapreduce(cos, *, collect(mx)),  mapreduce(cos, *, mx))
+    if VERSION >= v"1.4.0-DEV"
+        @inferred Union{Float64, Missing} mapreduce(cos, *, mx)
+    end
+
+    x = [missing missing missing]
+    y = [1, 2, 3]
+    mx, my = skipmissings(x, y)
+    @test_throws ArgumentError reduce(x -> x/2, mx)
+    @test_throws ArgumentError mapreduce(x -> x/2, +, mx)
+    @test_throws MethodError length(mx)
+    @test IndexStyle(typeof(mx)) == IndexStyle(typeof(x))
+    x = [isodd(i) ? missing : i for i in 1:64]
+    y = [isodd(i) ? missing : i for i in 65:128]
+    mx, my = skipmissings(x, y)
+    @test sum(mx) === 1056
+    @inferred Union{Missing, Int} sum(mx)
+    if VERSION >= v"1.4.0-DEV"
+        @inferred Union{Missing, Int} sum(mx)
+    end
+
     @test levels(1:1) == levels([1]) == levels([1, missing]) == levels([missing, 1]) == [1]
     @test levels(2:-1:1) == levels([2, 1]) == levels([2, missing, 1]) == [1, 2]
     @test levels([missing, "a", "c", missing, "b"]) == ["a", "b", "c"]
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ This package provides additional functionality for working with `missing` values
 - `allowmissing` and `disallowmissing` to convert between `Vector{T}` and `Vector{Union{T, Missing}}`
 - `passmissing` to wrap a function so that it returns `missing` if any of its positional arguments is `missing`
 - `levels` to get the unique values in a vector excluding `missing` and in their preferred order
+- `skipmissings` to loop through a collection of iterators excluding indi ces where any iterators are `missing`
 
 ## Contributing and Questions
 


src/Missings.jl
```diff
@@ -207,4 +208,74 @@ missing
...
+   multiskipmissing(args...)
+
+
+Returns a tuple of iterators iterating over indices of elements `args`
+where all elements of `args` are not missing.
+
+# Examples
+```
+julia> x = [1, 2, missing, 4]; y = [1, 2, 3, missing];
+julia> tx, ty = multiskipmissing(x, y)
```
nalimilan: Need to copy the output, which is interesting for users BTW. Also add a line break before (basically, just copy/paste the real output).

src/Missings.jl
```diff
@@ -207,4 +208,74 @@ missing
...
+julia> for t in tx
```
nalimilan: Probably simpler to show `collect(t)`?

src/Missings.jl
```diff
@@ -207,4 +208,74 @@ missing
...
+    item, state = y
+    while any(ismissing, item)
+        y = iterate(itr.c, ntuple(_ -> first(state), width(itr)))
+        y === nothing && return nothing
+        item, state = y
+    end
+    return item[itr.I], state
+end
+
+IndexStyle(::Type{<:MultiSkipMissing{T}}) where {T} = IndexStyle(T)
```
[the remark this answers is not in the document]
pdeffebach: `skipmissing` does support indexing (you implemented it 10 months ago). The logic is that it throws an error if an element is missing.

Which is odd behavior to be fair but we should support it for consistency.

pdeffebach:
I responded to comments. 

I also ported the `mapreduce` implementation for `skipmissing` over to this PR. But the implementation relies on *all* the objects involved in the `SkipMissings` object to be indexable, which is not known to the compiler given the way the object is currently constructed. 

Ideally the type of `SkipMissings` is something like 

```
struct SkipMissings{T, C} where T where C ::Base.Iterators.Zip{<:Tuple}
    c::C
    I::Int
end
```

Then we use `itr.I` to peer into `itr.c` when needed to get information about the array of interest etc. 

And we can dispatch on the elements of the `Tuple` in the `Zip`, but I'm not sure how to do that.

nalimilan:
> Ideally the type of `SkipMissings` is something like
> 
> ```
> struct SkipMissings{T, C} where T where C ::Base.Iterators.Zip{<:Tuple}
>     c::C
>     I::Int
> end
> ```
> 
> Then we use `itr.I` to peer into `itr.c` when needed to get information about the array of interest etc.
> 
> And we can dispatch on the elements of the `Tuple` in the `Zip`, but I'm not sure how to do that.

I'm not sure what prevents you from doing this. Can you show what you have tried?

pdeffebach:
I have updated this PR with a new implementation. Now the `SkipMissings` object contains only two fields, a `Zip` and an `Int`. When getting values, I use some gymnastics based off of the type signature to get elements from the array of interest. This requires a few more accessor functions which complicate the code a bit so I'm not sure it's worth it. 

Additionally, I have updated the type signatures for the optimized `reduce` implementations. These are complicated pieces of code so if you don't think they are useful, let me know and we can worry about them in a later PR.

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+end
+
+struct SkipMissings{C, I}
+    z::C
+    k::Int
+end
+_inner_type(C, I) = C.parameters[1].parameters[I]
+_inner_itr(S::SkipMissings) = S.z.is[S.k]
+_width(itr::SkipMissings) = length(first(itr.z))
+_anymissing(itr::SkipMissings, i) = any(x -> x === missing, (y[i] for y in itr.z.is))
```
nalimilan: I think the compiler will optimize this much better. Have you done some benchmarking?
```suggestion
_anymissing(itr::SkipMissings, i) = any(x -> x[i] === missing, itr.z.is)
```

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+const ZipofArrays = Base.Iterators.Zip{<:Tuple{Vararg{AbstractArray}}}
+
+Base.IteratorSize(::Type{<:SkipMissings}) = Base.SizeUnknown()
+Base.IteratorEltype(::Type{SkipMissings{C, I}}) where {C, I} = Base.IteratorEltype(_inner_type(C, I))
```
nalimilan:
```suggestion
Base.IteratorEltype(::Type{SkipMissings{C, I}}) where {C, I} =
    Base.IteratorEltype(_inner_type(C, I))
```

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+Base.eltype(::Type{SkipMissings{C, I}}) where {C, I} = nonmissingtype(Base.eltype(_inner_type(C, I)))
```
nalimilan:
```suggestion
Base.eltype(::Type{SkipMissings{C, I}}) where {C, I} =
    nonmissingtype(eltype(_inner_type(C, I)))
```

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+
+Base.iterate(itr::SkipMissings) = Base.iterate(itr, 1)
```
nalimilan: Replace this with `state=1` in the signature below.

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+        y = iterate(itr.z, z_state)
+        y === nothing && return nothing
+        z_item, z_state = y
+    end
+    return z_item[itr.k], z_state[itr.k]
+end
+
+Base.IndexStyle(::Type{<:SkipMissings{C, I}}) where {C, I} = Base.IndexStyle(_inner_type(C, I))
+
+function Base.eachindex(itr::SkipMissings{C, I}) where C <: ZipofArrays where I
```
nalimilan:
```suggestion
function Base.eachindex(itr::SkipMissings{C}) where C <: ZipofArrays
```

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+    @inbounds Iterators.filter(eachindex(_inner_itr(itr))) do i
+        !_anymissing(itr, i)
+    end
+end
+
+function Base.keys(itr::SkipMissings{C, I}) where C <: ZipofArrays where I
```
nalimilan:
```suggestion
function Base.keys(itr::SkipMissings{C}) where C <: ZipofArrays
```

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+    Iterators.filter(keys(_inner_itr(itr))) do i
+        !_anymissing(itr, i)
+    end
+end
+
+function Base.getindex(itr::SkipMissings{C, I}, i) where C <: ZipofArrays where I
```
nalimilan:
```suggestion
function Base.getindex(itr::SkipMissings{C}, i) where C <: ZipofArrays
```

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+    _anymissing(itr, i) && throw(MissingException("the value at index $i is missing for some element"))
+    _inner_itr(itr)[i]
+end
+
+# Optimized mapreduce implementation
+# The generic method is faster when !(eltype(A) >: Missing) since it does not need
+# additional loops to identify the two first non-missing values of each block
+Base.mapreduce(f, op, itr::SkipMissings{C, I}) where C <: ZipofArrays where I =
```
nalimilan:
```suggestion
Base.mapreduce(f, op, itr::SkipMissings{C}) where C <: ZipofArrays =
```

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+    Base._mapreduce(f, op, Base.IndexStyle(_inner_itr(itr)), Base.eltype(itr.z) <: Tuple{Vararg{Union{Missing, Any}}} ? 
+                                            itr : _inner_itr(itr))
+
+function Base._mapreduce(f, op, ::IndexLinear, itr::SkipMissings{C, I}) where C <: ZipofArrays where I
```
nalimilan:
```suggestion
function Base._mapreduce(f, op, ::IndexLinear, itr::SkipMissings{C}) where C <: ZipofArrays
```

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+    something(Base.mapreduce_impl(f, op, itr, first(inds), last(inds)))
+end
+
+Base._mapreduce(f, op, ::IndexCartesian, itr::SkipMissings) = mapfoldl(f, op, itr)
+
+Base.mapreduce_impl(f, op, A::SkipMissings, ifirst::Integer, ilast::Integer) =
+    Base.mapreduce_impl(f, op, A, ifirst, ilast, Base.pairwise_blocksize(f, op))
+
+# Returns nothing when the input contains only missing values, and Some(x) otherwise
+@noinline function Base.mapreduce_impl(f, op, itr::SkipMissings{C, I},
```
nalimilan:
```suggestion
@noinline function Base.mapreduce_impl(f, op, itr::SkipMissings{C},
```

src/Missings.jl
```diff
@@ -207,4 +208,177 @@ missing
...
+                                       ifirst::Integer, ilast::Integer, blksize::Int) where C <: ZipofArrays where I
```
nalimilan:
```suggestion
                                       ifirst::Integer, ilast::Integer, blksize::Int) where C <: ZipofArrays
```

test/runtests.jl
```diff
@@ -75,6 +75,20 @@ struct CubeRooter end
...
+    mx, my = skipmissings(x, y)
+    @test eltype(mx) == Int
+    @test collect(mx) == [1, 2]
+    @test collect(mx) isa Vector{Int} 
+    mx, my, mz = skipmissings(x, y, z)
+    @test isempty(collect(mx))
+    x = [missing 4; 2 5; 3 6]
+    y = [1 4; missing 5; 3 6]
+    mx, my = skipmissings(x, y)
+    @test collect(mx) == [3, 4, 5, 6]
```
nalimilan: Also test `mapreduce` (see tests for `skipmissing` in Julia, as these are tricky to cover 100%) and `getindex`. Put a few `@inferred` when it makes sense as that's really essential.

Would also be good to test `IndexStyle` and `length`/`size`.

test/runtests.jl
```diff
@@ -75,6 +75,20 @@ struct CubeRooter end
     @test collect(x) == [1, 2, 4]
     @test collect(x) isa Vector{Int}
 
+    x = [1, 2, missing, 4]
+    y = [1, 2, 3, missing]
+    z = [missing, missing, 3, 4]
```
nalimilan:
```suggestion
    z = [missing, missing, 3.1, 4.5]
```
And then test the different eltypes.

pdeffebach:
Thanks for the feedback! Two issues:

First, I don't understand how we can get rid of the `I` in the type signature of `SkipMissings`. Otherwise we don't know where to look for `eltype` etc. when all we have is the type `SkipMissings` and not an instance of the type. 

Second, we don't have inference when compared to `skipmissing` in Base. 

```
julia> x = [1, 2, missing, 4];
julia> y = ["a", "b", "c", missing];
julia> sx = skipmissing(x);
julia> mx, my = skipmissings(x, y);
julia> @code_warntype sum(sx)
# Variables
#   #self#::Core.Compiler.Const(sum, false)
#   a::Base.SkipMissing{Array{Union{Missing, Int64},1}}
#
# Body::Union{Missing, Int64}
# 1 ─ %1 = Base.sum(Base.identity, a)::Union{Missing, Int64}
# └──      return %1

julia> @code_warntype sum(mx)
# Variables
#   #self#::Core.Compiler.Const(sum, false)
#   a::Missings.SkipMissings{Base.Iterators.Zip{Tuple{Array{Union{Missing, Int64},1},Array{Union{Missing, String},1}}},1}
# 
# Body::Any
# 1 ─ %1 = Base.sum(Base.identity, a)::Any
# └──      return %1
```

I'm not sure how to fix it. Is the issue with my `const ZipofArrays` alias? Sorry that I wasn't able to figure this out.

nalimilan:
OK, I see. The problem is likely that `k` is just a value, not a type parameter, so it's not known at compile time. I think you either have to use `I` instead of `k`, or store the element type as a type parameter. The latter is probably better as it will avoid recompiling functions for each value of `k`, but probably better check the performance of both approaches. If that's not enough, you can add a type assertion to the first value returned by `iterate` (just like you did to `getindex`).

Something that needs to be checked is what happens when you iterate over all iterators returned by `skipmissings` (e.g. using `for (x, y) in zip(skipmissings(args...)...)` or `cor(skipmissings(args...)...)`): for performance it is essential that the compiler is able to avoid doing the iteration multiple times for each input. Maybe a special method for `SkipMissings{Zip{Tuple{<:AbstractVector}}}` with `@inbounds` will be needed for that.

pdeffebach:
Thanks for the suggestions, and sorry for taking so long to get back to this. 

i have added the index of the relevant vector as a type parameter and added lots of type asserts. Still no luck. Let me know if you have any new ideas. 

Otherwise I suppose we can iterate the vector separately. But that might have it's own performance drawbacks. 

I understand your other point and will work on it soon.

src/Missings.jl
```diff
@@ -237,29 +237,30 @@ julia> collect(tx)
...
 
 Base.IteratorSize(::Type{<:SkipMissings}) = Base.SizeUnknown()
-Base.IteratorEltype(::Type{SkipMissings{Z, I}}) where {Z, I} = Base.IteratorEltype(_inner_type(Z, I))
-Base.eltype(::Type{SkipMissings{Z, I}}) where {Z, I} = nonmissingtype(Base.eltype(_inner_type(Z, I)))
+Base.IteratorEltype(::Type{SkipMissings{Z, I, T}}) where {Z, I, T} = Base.IteratorEltype(T)
+Base.eltype(::Type{SkipMissings{Z, I, T}}) where {Z, I, T} = nonmissingtype(Base.eltype(T))
 
-function Base.iterate(itr::SkipMissings, state = 1)
-    y = iterate(itr.z, ntuple(_ -> state, _width(itr)))
+function Base.iterate(itr::SkipMissings{Z, I, T}, state = 1)::Tuple{eltype(T), Union{Int, Nothing}} where {Z, I, T}
```
nalimilan: `@code_warntype sum(mx)` from your previous post seems to be OK if I remove this failing type assert:
```suggestion
function Base.iterate(itr::SkipMissings{Z, I, T}, state = 1) where {Z, I, T}
```
or if I fix it:
```suggestion
function Base.iterate(itr::SkipMissings{Z, I, T}, state = 1)::Union{Tuple{eltype(T), Int}, Nothing} where {Z, I, T}
```

pdeffebach:
Still failing for me. I'm on 1.3 still. I will download the 1.5 and see if that's the issue.

pdeffebach:
The code correctly infers on 1.4-rc1! Now I will work on the `zip` propblem. 

I would assume we have a "less performant" release for 1.0-1.3 and then people in 1.4 just get the added benefit of having this code work better.

pdeffebach:
> Maybe a special method for `SkipMissings{Zip{Tuple{<:AbstractVector}}}` with `@inbounds` will be needed for that.

I understand the `Zip` argument and how that could be expensive. But can you describe what kind of improvement we can do for `cor(mx, my)` and similar functions? Looking at the code in `Statistics`,  `cor` doesn't use `zip` at all. 

To be fair, `Skipmissing` objects can't be used with `cor` because that just takes abstract vectors. Maybe we need some sort of `collect` function that works with many `skipmissings` objects?

nalimilan:
> The code correctly infers on 1.4-rc1! Now I will work on the `zip` propblem.
> 
> I would assume we have a "less performant" release for 1.0-1.3 and then people in 1.4 just get the added benefit of having this code work better.

OK, cool. Then yes, let's not worry about performance on older releases, at least for now. Pplease remove any type asserts that are not needed on Julia 1.4 (and anyway they are not enough on older releases).

> I understand the `Zip` argument and how that could be expensive. But can you describe what kind of improvement we can do for `cor(mx, my)` and similar functions? Looking at the code in `Statistics`, `cor` doesn't use `zip` at all.

By `Zip` I didn't mean that we should optimize `zip`, just that we could/should optimize `SkipMissings` when it gets passed only `AbstractArray`s (the most common case). But actually I'm not sure what's needed, so better first benchmark the overhead of `skipmissings`, e.g. compared to passing vectors directly when they don't contain missing values at all. Maybe it's going to be fast already.

There are two different paths to optimize: `iterate`, and `eachindex`+`getindex`. For `iterate`, it would be possible to override the generic definition with a more specific one for `AbstractArray` arguments, which would do `for i in eachindex(itr)`, and would then be able to use `@inbounds` when indexing the wrapped array. I'm not sure that would help, but it could be worth trying.

For the `eachindex`+`getindex` path, it will be a good idea anyway to put`@inline` before the `getindex` definition, and inside it check the bounds manually in a `@boundscheck` block, plus add `@inbounds` everywhere the index is used after that. `eachindex` already has `@inbounds`, but you must ensure that it's propagated to `_anymissing`, so add `@inline` to it too. See https://docs.julialang.org/en/latest/devdocs/boundscheck/ for details.

> To be fair, `Skipmissing` objects can't be used with `cor` because that just takes abstract vectors. Maybe we need some sort of `collect` function that works with many `skipmissings` objects?

Ah, indeed. We can probably extend that since the two objects barely need to be iterable and have the same length. Anyway, for now you can just test performance with a custom function, like this:
```julia
function testperfzip(X, Y)
    r = zero(eltype(x)) * zero(eltype(x))
    for (x, y) in zip(skipmissings(X, Y)...)
        r += x * y
    end
    r
end
```

src/Missings.jl
nalimilan: Don't forget this when benchmarking as it could make a difference.

pdeffebach:
I benchmarked this a bit yesterday and it turns out it's enormously slow. I will spend some time next week uncovering the source of the allocations. I think the reason has to do with `iterate` allocating and not something like `eachindex` and `@inbounds` type improvements.

pdeffebach:
Good news! I think this is about on par with what I was expecting, after some re-factoring. 

```
x = map(1:100) do _
    rand() < .2 ? missing : rand()
end

y = map(1:100) do _
    rand() < .2 ? missing : rand()
end

mx, my = skipmissings(x, y)
cx = collect(mx)
cy = collect(my)

function testperfzip(X, Y)
    r = zero(eltype(x)) * zero(eltype(x))
        for (x, y) in zip(X, Y)
            r += x * y
         end
         r
end

julia> @btime testperfzip($cx, $cy)
  1.692 μs (139 allocations: 2.17 KiB)
19.163942918070717

julia> @btime testperfzip($mx, $my)
  33.437 μs (579 allocations: 11.20 KiB)
19.163942918070717
```

nalimilan:
Cool. At least it looks like the overhead is reasonable. But have you tried with a larger vector? It's not clear whether the number of allocations increases with the length (it shouldn't).

Also, it would be interesting to check with vectors which don't contain any missing values: there the overhead should be low or non-existent, and that's a first goal to reach. Can you push your latest changes?

pdeffebach:
I think this is ready for another review. 

My performance improvement was a shift from using `Zip` internally to just using a `Tuple` of all the *other* iterators and writing some helper functions which perform better when checking if anything is missing. 

There is one big performance bottleneck and that is my use of 

```
const SkipMissingsofArrays = SkipMissings{V, T} where {V <: AbstractArray, T <: Tuple{Vararg{AbstractArray}}}
```

Which I then use to do the type signature of functions that need arrays as inputs, in particular `mapreduce`. Julia isn't doing what I am expecting it to do here. I have 

```
julia> mx isa Missings.SkipMissingsofArrays
julia> true
```

but `mapreduce` defaults to `any`, so something isn't working. Let me know if you have any thoughts.

nalimilan:
> Which I then use to do the type signature of functions that need arrays as inputs, in particular `mapreduce`. Julia isn't doing what I am expecting it to do here. I have
> 
> ```
> julia> mx isa Missings.SkipMissingsofArrays
> julia> true
> ```
> 
> but `mapreduce` defaults to `any`, so something isn't working. Let me know if you have any thoughts.

Have you checked that `Base.IndexStyle` returns `IndexLinear()`?

pdeffebach:
I added a method for `Base.IndexStyle`. It works now! I'm not 100% sure that's what made it work or there was some other error but it works now. 

In fact, `@code_warntype` works better than `skipmissing`. 

I think this PR is ready for a final review and then I would like to either 

1. Overload `zip` for `SkipMissings` so that we don't have to perform a double check every time we iterate when two arguments are together, which is a common occurence. 
2. Create a `SkipMissingsZip` that returns an iterator of tuples. 

Both can be done in a separate PR.

pdeffebach:
Nevermind. It was working and now it isn't. I'm not sure what changed and need to investigate further.

pdeffebach:
@nalimilan I think I need you to take a look at it, or I should ask Discourse / Slack for help because I'm not understanding the lack of inference. I believe the issue is with `Base.mapreduce_impl` defined for `SkipMissings`. 

The error has something to do with `Some(something(v))` and `Some` is where the type inference breaks, I think.

pdeffebach:
posted on discourse [here](https://discourse.julialang.org/t/help-with-type-inference/35753/2)

pdeffebach:
[Here](https://gist.github.com/pdeffebach/fd55c965b59ac783944d00e01bd8ace3) is a gist for the new performance behavior. Performance still isn't great, I think. Since `sum` is good, `iterate` is working okay. But `zip` is slow, meaning there could be mileage out of a customized `zip` implementation. 

```
Sum:
==========
N = 100: skipmissings
3.4100000000e-07 seconds
N = 100: vector
9.1581581582e-09 seconds
N = 10_000: skipmissings
3.5581000000e-05 seconds
N = 10_000: vector
6.5847468354e-07 seconds

Zip
==========
N = 100: skipmissings
2.9895000000e-05 seconds
N = 100: vector
5.5900919305e-08 seconds
N = 10_000: skipmissings
3.0154150000e-03 seconds
N = 10_000, vector
7.5552500000e-06 seconds
```

src/Missings.jl
```diff
@@ -207,4 +208,236 @@ missing
...
+end
+
+function Base.getindex(itr::SkipMissingsofArrays, i)::eltype(itr) 
+    vi = itr.x[i]
+    vi !== missing &&_anymissingindex(itr.others, i) && throw(MissingException("the value at index $i is missing for some element"))
+    vi
+end
+
+# Optimized mapreduce implementation
+# The generic method is faster when !(eltype(A) >: Missing) since it does not need
```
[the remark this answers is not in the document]
pdeffebach: deleting it. There is an optimization in `skipmissing` for if the array has no missing values that isn't really relevant here.

src/Missings.jl
```diff
@@ -207,4 +208,236 @@ missing
...
+    end
+end
+
+function Base.keys(itr::SkipMissingsofArrays)
+    Iterators.filter(keys(itr.x)) do i
+        itr.x[i] !== missing && !_anymissingindex(itr.others, i)
+    end
+end
+
+function Base.getindex(itr::SkipMissingsofArrays, i)::eltype(itr)
```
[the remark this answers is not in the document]
pdeffebach: no.

src/Missings.jl
```diff
@@ -207,4 +208,236 @@ missing
...
+        _first(iterate(oth, state)) === missing && return true
+    end
+
+    return false
+end
+
+const SkipMissingsofArrays = SkipMissings{V, T} where {V <: AbstractVector, T <: Tuple{Vararg{AbstractVector}}}
+
+function Base.show(io::IO, mime::MIME"text/plain", itr::SkipMissings) 
+    show(io, mime, "SkipMissings based $(itr.x) and $(length(itr.others)) other iterators")
```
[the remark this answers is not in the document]
pdeffebach: Not sure I understand this as I think that's what I'm doing. You are right that the printing is ugly right now but i'm not sure how to make it more compact.

pdeffebach:
I have responded to all comments above. 

* With regards to branching and inference, I settled on:

```
ai === missing || @inbounds _anymissingindex(itr.others, i) || (v = op(v, f(ai)))
```

which is hacky but infers. On the other hand it isn't any faster than nested `if` statements so let me know if I should change it back. Nested `if` statements in general seem just as fast as the short-circuited expressions. `sum` is still 30x slower than a sum with a `collect`ed vector, which is frustrating. 

* I will need your help on `show` because I'm not sure how to get it to be compact enough. 

* I added `@inbounds` in front of every `while` and `for` loop and deleted the `@inbounds` inside those loops. Performance is the same so I think I did it right.

src/Missings.jl
```diff
@@ -207,4 +208,256 @@ missing
...
+    @inbounds xi = itr.x[i]
+    if xi === missing || @inbounds _anymissingindex(itr.others, i) 
+        throw(MissingException("the value at index $i is missing for some element"))
+    end
+    return xi
+end
+
+function Base.mapreduce(f, op, itr::SkipMissingsofArrays)
+    Base._mapreduce(f, op, Base.IndexStyle(itr), itr)
+end
```
nalimilan:
```suggestion
Base.mapreduce(f, op, itr::SkipMissingsofArrays) =
    Base._mapreduce(f, op, Base.IndexStyle(itr), itr)
```

src/Missings.jl
```diff
@@ -207,4 +208,256 @@ missing
...
+ 9
+
+```
+"""
+function filter(f, itr::SkipMissingsofArrays)
+    x = itr.x
+    y = similar(x, eltype(itr), 0)
+    @inbounds for i in eachindex(x)
+        xi = x[i]
+        if xi !== missing && !_anymissingindex(itr.others, i) && f(xi)
```
nalimilan: Strictly speaking `@inbounds` should not be put here as it would apply to `f(xi)`, which could incorrectly assume some unrelated indices are in bounds.
```suggestion
    for i in eachindex(x)
        @inbounds xi = x[i]
        if xi !== missing && @inbounds !_anymissingindex(itr.others, i) && f(xi)
```

test/runtests.jl
nalimilan: This still applies AFAICT?

src/Missings.jl
```diff
@@ -207,4 +208,256 @@ missing
...
+function skipmissings(args...)
+    if args isa Tuple{Vararg{AbstractArray}}
+        if !mapreduce(length, isequal, args)
+            throw(ArgumentError("All arguments must have the same length"))
+        end
+
+        if !mapreduce(eachindex, isequal, args)
+            throw(ArgumentError("All arguments must have the same indices"))
+        end
+    end
```
nalimilan:
```suggestion
        if !mapreduce(length, ==, args)
            throw(ArgumentError("All arguments must have the same length"))
        end

        if !mapreduce(eachindex, ==, args)
            throw(ArgumentError("All arguments must have the same indices"))
        end
    end
```

src/Missings.jl
```diff
@@ -207,4 +208,256 @@ missing
...
+        SkipMissings(args[i], args[s])
+    end
+end
+
+struct SkipMissings{V, T}
+    x::V
+    others::T
+end
+
+Base.@propagate_inbounds @inline function _anymissingindex(others::Tuple{Vararg{AbstractArray}}, i)
```
nalimilan: I think `@inline` is redundant.
```suggestion
Base.@propagate_inbounds function _anymissingindex(others::Tuple{Vararg{AbstractArray}}, i)    
```

src/Missings.jl
```diff
@@ -207,4 +208,256 @@ missing
...
+    for oth in others
+        oth[i] === missing && return true
+    end
+
+    return false
+end
+
+_first(x) = first(x)
+_first(n::Nothing) = nothing
+
```
nalimilan:
```suggestion
```

src/Missings.jl
```diff
@@ -207,4 +208,256 @@ missing
...
+@inline function _anymissingiterate(others::Tuple, state)
+    for oth in others 
+        y = iterate(oth, state)
+        y !== nothing && first(y) === missing && return true
+    end
+
+    return false
+end
+
+const SkipMissingsofArrays = SkipMissings{V, T} where {V <: AbstractArray, T <: Tuple{Vararg{AbstractArray}}}
```
nalimilan:
```suggestion
const SkipMissingsofArrays = SkipMissings{V, T} where
    {V <: AbstractArray, T <: Tuple{Vararg{AbstractArray}}}
```

src/Missings.jl
```diff
@@ -207,4 +208,256 @@ missing
...
+    Base.mapreduce_impl(f, op, A, ifirst, ilast, Base.pairwise_blocksize(f, op))
+
+# Returns nothing when the input contains only missing values, and Some(x) otherwise
+@noinline function Base.mapreduce_impl(f, op, itr::SkipMissingsofArrays,
+                                       ifirst::Integer, ilast::Integer, blksize::Int)
+    A = itr.x
+    if ifirst == ilast
+        @inbounds a1 = A[ifirst]
+        a1 === missing || _anymissingindex(itr.others, ifirst) || return Some(Base.mapreduce_first(f, op, a1))
+        return nothing
```
nalimilan: Actually the previous version seemed more readable to me. And this one goes beyond 92 chars. Did the previous version infer correctly?
```suggestion
        if a1 === missing
            return nothing
        elseif _anymissingindex(itr.others, ifirst)
            return nothing
        else
            return Some(Base.mapreduce_first(f, op, a1))
        end
```

src/Missings.jl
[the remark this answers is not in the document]
pdeffebach: Okay now it says 

```
SkipMissings{...} Types... [array, array], comprised of 2 iterators
```

This is great. thanks for showing me how to do it.

src/Missings.jl
pdeffebach: Performance is the same. I changed it back to some `if... elseif` statements.

test/runtests.jl
pdeffebach: `mapreduce`, `getindex`, and `IndexStyle` were already covered. 

`length` and `size` are not supported so we don't test. 

I added an `@inferred` test.

src/Missings.jl
```diff
@@ -402,7 +406,7 @@ Base.mapreduce_impl(f, op, A::SkipMissingsofArrays, ifirst::Integer, ilast::Inte
         v = op(f(a1), f(a2))
         @simd for i = i:ilast
             @inbounds ai = A[i]
-            ai === missing || @inbounds !_anymissingindex(itr.others, i) || (v = op(v, f(ai)))
+            ai === missing || @inbounds _anymissingindex(itr.others, i) || (v = op(v, f(ai)))
```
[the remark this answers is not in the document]
pdeffebach: added a test for this.

src/Missings.jl
```diff
@@ -207,4 +208,256 @@ missing
...
+
+function Base.iterate(itr::SkipMissings{V, T}, state = 0) where {V <: AbstractArray, T <: Tuple{Vararg{AbstractArray}}}
+    eix = eachindex(itr.x)
+    inditr = iterate(eix, state)
+    inditr === nothing && return nothing
+    ind, indstate = inditr
+    @inbounds x_item = itr.x[ind]
+    @inbounds while true
+        x_item === missing || _anymissingindex(itr.others, ind) || break
+        indnext = iterate(eix, indstate)
```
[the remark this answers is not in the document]
pdeffebach: fixed.

pdeffebach:
Thanks for the review. I have changed everything, including adding an `@inferred` test and making the code more readable.

src/Missings.jl
```diff
@@ -237,11 +234,11 @@ julia> collect(tx)
 """
 function skipmissings(args...)
```
nalimilan: Throw an error if `isempty(args)`?

src/Missings.jl
```diff
@@ -212,21 +212,18 @@ passmissing(f) = PassMissing{Core.Typeof(f)}(f)
...
 
 julia> tx, ty = skipmissings(x, y)
-(Missings.SkipMissings{Array{Union{Missing, Int64},1}}(Union{Missing, Int64}[1, 2, missing, 
-4], Base.Iterators.Zip{Tuple{Array{Union{Missing, Int64},1},Array{Union{Missing, Int64},1}}}
-((Union{Missing, Int64}[1, 2, missing, 4], Union{Missing, Int64}[1, 2, 3, missing])), 1), 
-Missings.SkipMissings{Array{Union{Missing, Int64},1}}(Union{Missing, Int64}[1, 2, 3, 
-missing], Base.Iterators.Zip{Tuple{Array{Union{Missing, Int64},1},Array{Union{Missing,
- Int64},1}}}((Union{Missing, Int64}[1, 2, missing, 4], Union{Missing, Int64}[1, 2, 3, 
- missing])), 2))
+(SkipMissings{Array{Union{Missing, Int64},1},Tuple{Array{Union{Missing, Int64},1}}}(Union{Mi
```
nalimilan: This doesn't seem to match the new `show`, does it?

src/Missings.jl
```diff
@@ -406,7 +405,13 @@ Base.mapreduce_impl(f, op, A::SkipMissingsofArrays, ifirst::Integer, ilast::Inte
         v = op(f(a1), f(a2))
         @simd for i = i:ilast
             @inbounds ai = A[i]
-            ai === missing || @inbounds _anymissingindex(itr.others, i) || (v = op(v, f(ai)))
+            if ai === missing
```
nalimilan: This looks really weird. Was there a problem with the previous version using `||`? (Sorry if my comment about the other block prompted you to change this one too. I think the fact that the other has `return` makes the situation different.)

test/runtests.jl
```diff
@@ -107,6 +107,12 @@ struct CubeRooter end
     @test_throws ArgumentError mapreduce(x -> x/2, +, mx)
     @test_throws MethodError length(mx)
     @test IndexStyle(typeof(mx)) == IndexStyle(typeof(x))
+    x = [isodd(i) ? missing : i for i in 1:64]
+    y = [isodd(i) ? missing : i for i in 65:128]
+    mx, my = skipmissings(x, y)
+    @test sum(mx) == 1056
+    @test Union{Missing, Int} sum(mx)
```
nalimilan: ?

src/Missings.jl
```diff
@@ -1,7 +1,8 @@
 module Missings
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
-       Missing, MissingException, levels, coalesce, passmissing, nonmissingtype
+       Missing, MissingException, levels, coalesce, passmissing, nonmissingtype,
+       skipmissings, SkipMissings
```
nalimilan: Base doesn't export `SkipMissing`, so let's do the same here (this will affect the result of `show`, so be careful to update it).
```suggestion
       skipmissings
```

src/Missings.jl
```diff
@@ -207,4 +208,262 @@ missing
...
+julia> x = [1, 2, missing, 4]; y = [1, 2, 3, missing];
+
+julia> tx, ty = skipmissings(x, y)
+(SkipMissings{Array{Union{Missing, Int64},1},Tuple{Array{Union{Missing, Int64},1}}}(Union{Mi
+ssing, Int64}[1, 2, missing, 4], (Union{Missing, Int64}[1, 2, 3, missing],)), SkipMissings{A
+rray{Union{Missing, Int64},1},Tuple{Array{Union{Missing, Int64},1}}}(Union{Missing, Int64}[1
+, 2, 3, missing], (Union{Missing, Int64}[1, 2, missing, 4],)))
+
+julia> collect(tx)
+2-element Array{Any,1}:
```
nalimilan: Does this really produce an `Array{Any}`? If so that's a bug. Can you add a test for that?

test/runtests.jl
```diff
@@ -75,6 +75,45 @@ struct CubeRooter end
...
 
+    x = [1, 2, missing, 4]
+    y = ["a", "b", "c", missing]
+    z = [missing, missing, 3.1, 4.5]
+    l = [1, 2, 3, 4, 5]
+    @test_throws ArgumentError skipmissings(x, l)
+    mx, my = skipmissings(x, y)
+    @test collect(mx) == [1, 2]
+    @test collect(mx) isa Vector{Int} 
+    @test reduce(+, mx) == reduce(+, collect(mx)) == sum(mx) ==
```
nalimilan: Can you also test the type of the result?

test/runtests.jl
```diff
@@ -75,6 +75,45 @@ struct CubeRooter end
...
+        mapreduce(identity, +, mx) == 3
+    @test mapreduce(x -> x^2, +, mx) == mapreduce(x -> x^2, +, collect(mx)) == 5
+    mx, my, mz = skipmissings(x, y, z)
+    @test eltype(mx) == Int
+    @test eltype(my) == String
+    @test eltype(mz) <: AbstractFloat
```
nalimilan:
```suggestion
    @test eltype(mz) == Float64
```

test/runtests.jl
```diff
@@ -75,6 +75,45 @@ struct CubeRooter end
...
+    @test isempty(collect(mx))
+    @test sum(mx) == 0
+    x = [missing 4; 2 5; 3 6]
+    y = [1 4; missing 5; 3 6]
+    mx, my = skipmissings(x, y)
+    @test collect(mx) == [3, 4, 5, 6]
+    @test mx[3] == 3
+    @test_throws MissingException mx[1]
+    @test reduce(+, mx) == 18
```
nalimilan:
```suggestion
    @test reduce(+, mx) === 18
```

test/runtests.jl
```diff
@@ -75,6 +75,45 @@ struct CubeRooter end
...
+    @test collect(mx) isa Vector{Int} 
+    @test reduce(+, mx) == reduce(+, collect(mx)) == sum(mx) ==
+        mapreduce(identity, +, mx) == 3
+    @test mapreduce(x -> x^2, +, mx) == mapreduce(x -> x^2, +, collect(mx)) == 5
+    mx, my, mz = skipmissings(x, y, z)
+    @test eltype(mx) == Int
+    @test eltype(my) == String
+    @test eltype(mz) <: AbstractFloat
+    @test isempty(collect(mx))
+    @test sum(mx) == 0
```
nalimilan:
```suggestion
    @test sum(mx) === 0
```

test/runtests.jl
```diff
@@ -75,6 +75,45 @@ struct CubeRooter end
...
+    x = [missing 4; 2 5; 3 6]
+    y = [1 4; missing 5; 3 6]
+    mx, my = skipmissings(x, y)
+    @test collect(mx) == [3, 4, 5, 6]
+    @test mx[3] == 3
+    @test_throws MissingException mx[1]
+    @test reduce(+, mx) == 18
+    @test isapprox(mapreduce(cos, *, collect(mx)),  mapreduce(cos, *, mx))
```
nalimilan: Can you also test the return type here?

src/Missings.jl
pdeffebach: it matches.

src/Missings.jl
pdeffebach: sorry, no problem. Using the `||` for this line, then.

pdeffebach: Actually, let me confirm that you see we are assigning a variable `v` after the last `||`. 

It feels weird to assign a variable at the end of all these short-circuited booleans.

src/Missings.jl
pdeffebach: It infers correctly, thankfully.

nalimilan: But it shouldn't produce an `Array{Any}`. Why not an `Array{Int}`?

src/Missings.jl
nalimilan: Yeah it's a bit complex. Normally I would have written this as:
```julia
if ai !== missing && @inbounds !_anymissingindex(itr.others, i)
    v = op(v, f(ai))
end
```
But if the compiler doesn't like it, I'd rather use `||` than empty `if` branches.

src/Missings.jl
nalimilan: Where did "comprised of" go then? :-D

src/Missings.jl
pdeffebach: nevermind (sorry for spamming you with this). There is a very slight performance improvement to using the `||` version so I am keeping it.

pdeffebach:
Thanks for the feedback!

Here is an up-to-date performance test:

```
============
Proportion missing: 0.2
============

Sum:
==========
N = 100: skipmissings
    2.9627500000e-07 seconds

N = 100: vector
    1.2062124248e-08 seconds

N = 100: manual sum with view
    5.8031318681e-07 seconds

N = 100: manual sum with index
    4.6250520833e-07 seconds

N = 10_000: skipmissings
    2.9793000000e-05 seconds

N = 10_000: vector
    1.1262000000e-06 seconds

N = 10_000: manual sum with view
    2.5565000000e-05 seconds

N = 10_000: manual sum with index
    2.3489000000e-05 seconds


Zip
==========
N = 100: skipmissings
    7.0433566434e-07 seconds

N = 100: vector
    1.0167311828e-07 seconds

N = 100: manual skipping missings
    2.2379253112e-07 seconds

N = 10_000: skipmissings
    1.0437200000e-04 seconds

N = 10_000, vector
    1.1833000000e-05 seconds

N = 10_000: manual skipping missings
    4.6975000000e-05 seconds
```

The `Zip` sum is about twice the speed as a manual sum skippimg missings. However this will be fixed with an optimized implementation of `Zip` for skipmissings.

src/Missings.jl
pdeffebach: You are right. The `show` method in this docstring is for the `Tuple` of `(tx, ty)`... Just printing `tx` in the REPL gives the right answer, but the `Tuple` printing overrides it.

nalimilan: Ah, indeed. I saw just added a commit to fix this, but actually given that https://github.com/JuliaLang/julia/pull/34734 recommends, 2-argument `show` to print a parseable output, we should probably leave it that way even if it's not so nice.

src/Missings.jl
```diff
@@ -207,4 +208,263 @@ missing
...
+    end
+
+    return false
+end
+
+const SkipMissingsofArrays = SkipMissings{V, T} where
+    {V <: AbstractArray, T <: Tuple{Vararg{AbstractArray}}}
+
+function Base.show(io::IO, mime::MIME"text/plain", itr::SkipMissings{V}) where V 
+    print(io, SkipMissings, '{', V, '}', '(', itr.x, ')', " comprised of $(length(itr.others) + 1) iterators")
```
nalimilan: Break at 92 chars.

pdeffebach:
Thanks again for the feedback. All ready.

src/Missings.jl
```diff
@@ -278,11 +279,8 @@ const SkipMissingsofArrays = SkipMissings{V, T} where
     {V <: AbstractArray, T <: Tuple{Vararg{AbstractArray}}}
 
 function Base.show(io::IO, mime::MIME"text/plain", itr::SkipMissings{V}) where V 
-    print(io, SkipMissings, '{', V, '}', '(', itr.x, ')', " comprised of $(length(itr.others) + 1) iterators")
-end
-
-function Base.show(io::IO, itr::SkipMissings{V}) where V
-    print(io, SkipMissings, '{', V, '}', '(', itr.x, ')', " comprised of $(length(itr.others) + 1) iterators")
+    print(io, SkipMissings, '{', V, '}', '(', itr.x, ')', " comprised of " *
+        "$(length(itr.others) + 1) iterators")
```
nalimilan:
```suggestion
    print(io, SkipMissings, '{', V, '}', '(', itr.x, ')', " comprised of " *
          "$(length(itr.others) + 1) iterators")
```

test/runtests.jl
nalimilan: Still applies.

test/runtests.jl
```diff
@@ -75,6 +75,45 @@ struct CubeRooter end
...
+    y = ["a", "b", "c", missing]
+    z = [missing, missing, 3.1, 4.5]
+    l = [1, 2, 3, 4, 5]
+    @test_throws ArgumentError skipmissings(x, l)
+    mx, my = skipmissings(x, y)
+    @test collect(mx) == [1, 2]
+    @test collect(mx) isa Vector{Int} 
+    @test reduce(+, mx) == reduce(+, collect(mx)) == sum(mx) ==
+        mapreduce(identity, +, mx) == 3
+    @test mapreduce(x -> x^2, +, mx) == mapreduce(x -> x^2, +, collect(mx)) == 5
```
nalimilan:
```suggestion
    @test mapreduce(x -> x^2, +, mx) === mapreduce(x -> x^2, +, collect(mx)) === 5
```

test/runtests.jl
nalimilan: Still applies: use `===` instead of `==` to test type.

test/runtests.jl
```diff
@@ -75,6 +75,45 @@ struct CubeRooter end
...
+    @test_throws ArgumentError reduce(x -> x/2, mx)
+    @test_throws ArgumentError mapreduce(x -> x/2, +, mx)
+    @test_throws MethodError length(mx)
+    @test IndexStyle(typeof(mx)) == IndexStyle(typeof(x))
+    x = [isodd(i) ? missing : i for i in 1:64]
+    y = [isodd(i) ? missing : i for i in 65:128]
+    mx, my = skipmissings(x, y)
+    @test sum(mx) == 1056
+    @inferred Union{Missing, Int} sum(mx)
+
```
nalimilan: Also add a test for `show`.

test/runtests.jl
pdeffebach: it's added below with `@inferred`.

pdeffebach:
Finished the changes and successfully avoided a rebase. (I had changed a `passmissing` test which was fixed between opening this PR and now). The test now fails but it won't fail when we merge.

nalimilan:
Thanks. Can you rebase anyway? I don't want to take the risk of making CI fail on master (unexpected failures on rebase are frequent).

pdeffebach:
I believe I successfully rebased! There were no conflicts.

src/Missings.jl
nalimilan: I'm not sure what happened but this suggestion got reverted. Any idea why? Hopefully it's the only one.

pdeffebach:
As you know, my rebasing skills are weak. Googling around it seems like a re-base is too complicated for this scenario because I have too many fixes to code I added during the course of this PR. This is an alternate method I got off slack. 

It works when I try it on a test branch. 

```
git checkout multiskipmissing # this branch
git reset --soft <commit I diverged from>
git stash
git checkout master
git checkout -b temp_branch
git stash pop # will have to resolve conflicts in two lines. It's nbd. 
git commit -am "Squash everything"
git checkout multiskipmissing 
git reset --hard temp_branch # scary!
```

pdeffebach:
I propose to do this, but then push `temp_branch` to a new PR and merge it master in from there.

nalimilan:
I think the rebase is OK (except for that mysterious `@inline` change). But you need to make the `@inferred` test conditional on `VERSION` to fix CI.

src/Missings.jl
```diff
@@ -207,4 +208,261 @@ missing
 """
 passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
+
+"""
+   skipmissings(args...)
+
+Return a tuple of iterators wrapping each of the iterators in `args`, but
+skipping elements at positions where at least one of the iterators returns `missing`.
```
nalimilan, line 216: It just occurred to me that it could be useful to mention this keyword:
```suggestion
skipping elements at positions where at least one of the iterators returns `missing`
(listwise deletion of missing values).
```

pdeffebach:
The rebase didn't work all the way, I still think it's because of all the re-factoring that I did mid-PR. You can tell because the `passmissing` test still fails but would have not failed had the rebase gone through. But I updated the tests and docs.

pdeffebach:
I can confirm that `git merge master` works perfectly, though. with no conflicts.

nalimilan:
But the `passmissing` test doesn't seem to fail (CI passes everywhere except Julia 1.0)?

pdeffebach:
It doesn't seem to be testing 1.4 yet, so it passes. The PR to have it test 1.4 is not in here because the rebase didn't work. If I `merge` then it will see it.

nalimilan:
But master doesn't appear to test 1.4. I've pulled locally and rebased against latest master, but indeed the branch seems to be really messy. I think the simplest solution would be to squash everything before rebasing, so that you only have to fix conflicts once.

test/runtests.jl
```diff
@@ -75,6 +75,56 @@ struct CubeRooter end
...
+    mx, my = skipmissings(x, y)
+    @test_throws ArgumentError reduce(x -> x/2, mx)
+    @test_throws ArgumentError mapreduce(x -> x/2, +, mx)
+    @test_throws MethodError length(mx)
+    @test IndexStyle(typeof(mx)) == IndexStyle(typeof(x))
+    x = [isodd(i) ? missing : i for i in 1:64]
+    y = [isodd(i) ? missing : i for i in 65:128]
+    mx, my = skipmissings(x, y)
+    @test sum(mx) === 1056
+    @inferred Union{Missing, Int} sum(mx)
```
nalimilan, line 123:
```suggestion
```

nalimilan:
Argh, since the `@inferred` failure happens at parsing time, I think you need to wrap it in a `@eval` block.

pdeffebach:
Tests pass!

test/runtests.jl
```diff
@@ -75,6 +75,61 @@ struct CubeRooter end
...
+    y = ["a", "b", "c", missing]
+    z = [missing, missing, 3.1, 4.5]
+    l = [1, 2, 3, 4, 5]
+    @test_throws ArgumentError skipmissings(x, l)
+    mx, my = skipmissings(x, y)
+    iobuf = IOBuffer()
+    show(iobuf, MIME("text/plain"), mx)
+    s = String(take!(iobuf))
+    @test s == "Missings.SkipMissings{Array{Union{Missing, Int$(Sys.WORD_SIZE)},1}}(Union{Missing, Int$(Sys.WORD_SIZE)" *
+        "}[1, 2, missing, 4]) comprised of 2 iterators"
```
nalimilan, line 88:
```suggestion
    @test s == "Missings.SkipMissings{Array{Union{Missing, $Int},1}}(" *
        "Union{Missing, $Int}[1, 2, missing, 4]) comprised of 2 iterators"
```

test/runtests.jl
```diff
@@ -75,6 +75,61 @@ struct CubeRooter end
...
+    y = [1 4; missing 5; 3 6]
+    mx, my = skipmissings(x, y)
+    @test collect(mx) == [3, 4, 5, 6]
+    @test mx[3] == 3
+    @test_throws MissingException mx[1]
+    @test reduce(+, mx) === 18
+    @test isapprox(mapreduce(cos, *, collect(mx)),  mapreduce(cos, *, mx))
+    if VERSION >= v"1.4.0-DEV"
+        t = quote 
+            @inferred Union{Float64, Missing} mapreduce(cos, *, $mx)
```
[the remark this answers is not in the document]
pdeffebach: For the `skipmissing` in Base we have

```
julia> @code_warntype sum(sx)
Variables
  #self#::Core.Compiler.Const(sum, false)
  a::Base.SkipMissing{Array{Union{Missing, Int64},1}}

Body::Union{Missing, Int64}
1 ─ %1 = Base.sum(Base.identity, a)::Union{Missing, Int64}
└──      return %1
```

I don't know why the output works like that given `eltype(sx) == Int`  but it's a yet-to-be improved aspect of `mapreduce` with `skipmissing`. So it's fair that `skipmissings` would work the same. You have a comment on some issue highlighting that it's not a high priority performance issue but I can't find it. 

Thankfully small unions are fast.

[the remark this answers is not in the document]
nalimilan: OK, I need to investigate this. It's somewhat ironic that the function designed to skip missings isn't inferred as such. :-D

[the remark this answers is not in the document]
nalimilan: Actually I've realized `@static if` works here instead of `@eval`. I've committed that.

nalimilan:
Thanks!

PR #112, opened 2020-01-15
Fix passmissing exception test on Julia 1.4

nalimilan:


bkamins:
I understand that in Julia 1.4 passing a kwarg that is not supported is `MethodError` - right?

nalimilan:
Yes, that seems to have changed over the last 3 months.

PR #114, opened 2020-03-25
Define anymissing

arnavs:
See also https://github.com/JuliaLang/julia/pull/35237.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This package provides additional functionality for working with `missing` values
 - `allowmissing` and `disallowmissing` to convert between `Vector{T}` and `Vector{Union{T, Missing}}`
 - `passmissing` to wrap a function so that it returns `missing` if any of its positional arguments is `missing`
 - `levels` to get the unique values in a vector excluding `missing` and in their preferred order
-- `hasmissing` tells if you an iteratable has any `missing` values
+- `hasmissing` to see if an iteratable has any `missing` values
 
 ## Contributing and Questions
 
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ This package provides additional functionality for working with `missing` values
 - `allowmissing` and `disallowmissing` to convert between `Vector{T}` and `Vector{Union{T, Missing}}`
 - `passmissing` to wrap a function so that it returns `missing` if any of its positional arguments is `missing`
 - `levels` to get the unique values in a vector excluding `missing` and in their preferred order
+- `hasmissing` tells if you an iteratable has any `missing` values
 
 ## Contributing and Questions
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This package provides additional functionality for working with `missing` values
 - `allowmissing` and `disallowmissing` to convert between `Vector{T}` and `Vector{Union{T, Missing}}`
 - `passmissing` to wrap a function so that it returns `missing` if any of its positional arguments is `missing`
 - `levels` to get the unique values in a vector excluding `missing` and in their preferred order
-- `hasmissing` to see if an iteratable has any `missing` values
+- `hasmissing` to check whether an iterable has any `missing` values
 
 ## Contributing and Questions
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This package provides additional functionality for working with `missing` values
 - `allowmissing` and `disallowmissing` to convert between `Vector{T}` and `Vector{Union{T, Missing}}`
 - `passmissing` to wrap a function so that it returns `missing` if any of its positional arguments is `missing`
 - `levels` to get the unique values in a vector excluding `missing` and in their preferred order
-- `hasmissing` to check whether an iterable has any `missing` values
+- `anymissing` to check whether an iterable has any `missing` values
 
 ## Contributing and Questions
 
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -216,5 +216,6 @@ Return `true` if `itr` contains `missing` elements, else `false`.
 See also: `ismissing`
 """
 hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
+hasmissing(::Base.SkipMissing) = false 
 
 end # module
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -209,7 +209,7 @@ passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
 
 """
-    hasmissing(itr)
+    hasmissing(itr) -> Bool
 
 Return `true` if `itr` contains `missing` elements. 
 """
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -1,7 +1,7 @@
 module Missings
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
-       Missing, MissingException, levels, coalesce, passmissing, nonmissingtype
+       Missing, MissingException, levels, coalesce, passmissing, nonmissingtype, hasmissing
 
 using Base: ismissing, missing, Missing, MissingException
 using Base: @deprecate
@@ -206,5 +206,13 @@ julia> passmissing((x,y)->"\$x \$y")(missing)
 missing
 """
 passmissing(f) = PassMissing{Core.Typeof(f)}(f)
+
+
+"""
+    hasmissing(itr)
+
+Return `true` if `itr` contains `missing` elements. 
+"""
+hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
 
 end # module
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -211,7 +211,9 @@ passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 """
     hasmissing(itr) -> Bool
 
-Return `true` if `itr` contains `missing` elements. 
+Return `true` if `itr` contains `missing` elements, else `false`.
+
+See also: `ismissing`
 """
 hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
 
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -1,7 +1,7 @@
 module Missings
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
-       Missing, MissingException, levels, coalesce, passmissing, nonmissingtype, hasmissing
+       Missing, MissingException, levels, coalesce, passmissing, nonmissingtype, anymissing
 
 using Base: ismissing, missing, Missing, MissingException
 using Base: @deprecate
@@ -209,13 +209,13 @@ passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
 
 """
-    hasmissing(itr) -> Bool
+    anymissing(itr) -> Bool
 
 Return `true` if `itr` contains `missing` elements, else `false`.
 
 See also: `ismissing`
 """
-hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
-hasmissing(::Base.SkipMissing) = false 
+anymissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
+anymissing(::Base.SkipMissing) = false
 
 end # module
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -166,6 +166,11 @@ struct CubeRooter end
     @test passmissing(Int) === Missings.PassMissing{Type{Int}}(Int)
     @test passmissing(cuberoot) === Missings.PassMissing{CubeRooter}(cuberoot)
 
+    
+    @test hasmissing([1, 2, 3, 4]) == false
+    @test hasmissing([1, 2, 3, missing, 5]) == true
+    @test hasmissing(skipmissing([1, 2, 3, missing, 5])) == false
+    
     @testset "deprecated" begin
         # The (unexported) `Missings.T` was deprecated to `Missings.nonmissingtype`
         for x in (Union{Int, Missing}, Any, Missing, Union{Array{Int}, Missing})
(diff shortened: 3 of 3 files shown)


src/Missings.jl
```diff
@@ -207,4 +207,12 @@ missing
 """
 passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
+
+"""
+    hasmissing(itr)
```
nickrobinson251: May as well be as explicit as possible
```suggestion
    hasmissing(itr) -> Bool
```

src/Missings.jl
```diff
@@ -207,4 +207,12 @@ missing
...
+
+Return `true` if `itr` contains `missing` elements.
```
nickrobinson251:
```suggestion
Return `true` if `itr` contains `missing` elements, else `false`.

See also: `ismissing`
```

README.md
```diff
@@ -22,6 +22,7 @@ This package provides additional functionality for working with `missing` values
 - `allowmissing` and `disallowmissing` to convert between `Vector{T}` and `Vector{Union{T, Missing}}`
 - `passmissing` to wrap a function so that it returns `missing` if any of its positional arguments is `missing`
 - `levels` to get the unique values in a vector excluding `missing` and in their preferred order
+- `hasmissing` to see if an iteratable has any `missing` values
```
nalimilan:
```suggestion
- `hasmissing` to check whether an iterable has any `missing` values
```

nalimilan:
Can you show a case in which `any(ismissing, itr)` isn't as efficient as `hasmissing` from this PR?

nickrobinson251:
> Can you show a case in which `any(ismissing, itr)` isn't as efficient as `hasmissing` from this PR?

```julia
julia> using BenchmarkTools

julia> x = 1:100_000;

julia> hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
hasmissing (generic function with 1 method)

julia> @btime hasmissing(x);
  13.214 ns (0 allocations: 0 bytes)

julia> @btime any(ismissing, x);
  99.527 ns (2 allocations: 64 bytes)
```

nalimilan:
@nickrobinson251 You need to wrap the call in a function:
```julia
julia> using BenchmarkTools

julia> x = 1:100_000;

julia> hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
hasmissing (generic function with 1 method)

julia> @btime hasmissing(x);
  14.094 ns (0 allocations: 0 bytes)

julia> @btime any(ismissing, x);
118.323 ns (2 allocations: 64 bytes)

julia> f(x) = any(ismissing, x)
f (generic function with 1 method)

julia> @btime f(x);
  16.714 ns (0 allocations: 0 bytes)
```

arnavs:
I believe you also need to interpolate the argument into the macro. 

```
julia> x = [1, 2, 3, 4, 5, 6, 7]
7-element Array{Int64,1}:
 1
 2
 3
 4
 5
 6
 7

julia> hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
hasmissing (generic function with 1 method)

julia> f(x) = any(ismissing, x)
f (generic function with 1 method)

julia> @btime f($x)
  1.312 ns (0 allocations: 0 bytes)
false

julia> @btime hasmissing($x)
  0.027 ns (0 allocations: 0 bytes)
false
```

nalimilan:
When you reach nanoseconds, time differences may not be very significant. If you look at `@code_warntype`, you can see that in both cases the compiler statically replaced the call with `false`. So you really have to look at what difference it makes in a real application.

arnavs:
So shall we remove the `eltype(x) >: Missing`? Personally the logic makes sense to me, but it's possible that the compiler optimizes it away.

nalimilan:
No, what I'm saying is that it only makes sense to add `hasmissing` if we can find at least one situation in which it's clearly faster than `any(ismissing, x)`.

nickrobinson251:
Perhaps @bkamins can comment on a case where this did not optimize out? (ref https://github.com/JuliaLang/julia/pull/35237#discussion_r398162053)

arnavs:
I think it's just cleaner in many respects... for example `filter(hasmissing, some_datasets)`, or `hasmissing(x) && do_something`. 

Just makes the code easier to read... like convenience functions should.

bkamins:
> No, what I'm saying is that it only makes sense to add `hasmissing` if we can find at least one situation in which it's clearly faster than `any(ismissing, x)`.

Well you can do for example:

```
julia> x = Real[i for i in 1:10^6];

julia> @btime hasmissing($x);
  0.024 ns (0 allocations: 0 bytes)

julia> f(x) = any(ismissing, x)
f (generic function with 1 method)

julia> @btime f($x);
  522.209 μs (0 allocations: 0 bytes)
```

src/Missings.jl
```diff
@@ -207,4 +207,14 @@ missing
...
 
+
+"""
+    hasmissing(itr) -> Bool
+
+Return `true` if `itr` contains `missing` elements, else `false`.
+
+See also: `ismissing`
+"""
+hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
```
bkamins: You still need to define a special method for `skipmissing` (as in your original PR I have commented on) for performance reasons:
```
julia> x = Any[i for i in 1:10^6];

julia> y = skipmissing(x);

julia> hasmissing(itr) = eltype(itr) >: Missing && any(ismissing, itr)
hasmissing (generic function with 1 method)

julia> @btime hasmissing($y);
  8.464 ms (0 allocations: 0 bytes)
```

arnavs: True.

nalimilan:
> I think it's just cleaner in many respects... for example `filter(hasmissing, some_datasets)`, or `hasmissing(x) && do_something`.
> 
> Just makes the code easier to read, without any penalty in speed... like convenience functions should.

I'm not really convinced by the convenience argument. `any(ismissing, x) && do_something` is quite concise and clear, and it doesn't require people to look at `?hasmissing` to find out what it means exactly. Regarding the `filter` example, do you have a concrete use case in mind? Note that `hasmissing` doesn't work on DataFrames with this PR.

Regarding performance, the `Array{Real}` case is a good illustration, but that kind of array is inefficient already so I'm not sure it's worth optimizing for this (and the compiler could improve to cover this).

bkamins:
> and the compiler could improve to cover this

This is true (actually I expected the compiler to optimize it out but it did not), but the case of `any(ismissing, skipmissing(Any[1,2,3]))` is more common (when you read in potentially heterogeneous data) and not possible to be optimized out (unless we ad a method for `any(::type(ismissing), ::SkipMissing)` covering this specific case explicitly). Still - agreed that is not going to be a bottleneck probably. This is why I opted for `hasmissing` to live in Missings.jl and not in Base.

nalimilan:
> unless we ad a method for `any(::type(ismissing), ::SkipMissing)` covering this specific case explicitly

Actually we can do that, and probably should, as it doesn't introduce any ambiguity. :-)

arnavs:
So I think we are good to merge then?

nickrobinson251:
I think this is a nice little addition, that's faster in some otherwise very slow cases, and simplifies a common patern for some users (with `filter(hasmissing, some_datasets)` being much nicer than `filter(x -> any(ismissing, x) some_datasets)`). I think Missings.jl is a good home for it.

arnavs:
So is anything blocking this then?

bkamins:
> (Do we need @nalimilan to sign off on it?)

I would like to convince @nalimilan first. He has designed `Missing` type in Base.

arnavs:
@bkamins OK, sounds good.

arnavs:
We could also add:

```
hasmissing(df::DataFrame) = hasmissing(Array(df))
```

To make it applicable to those types.

bkamins:
It would rather be:
```
hasmissing(df::AbstractDataFrame) = any(hasmissing, eachcol(df))
```
and would live in DataFrames.jl not here. But let us first decide on the main use case.

(also note we already have `completecases` function which is very similar but gives you more information at the cost of higher computation cost)

arnavs:
Fair enough. Let's wait for @nalimilan to get back to us and @nickrobinson251 then.

nalimilan:
Actually the case of `DataFrame` is a better example of the potential use of this function, as you currently have to write `any(x -> any(ismissing, x), eachcol(df))`.

So I'd be on board with this for this reason. Regarding the name, what do you think about `anymissing` instead?

nickrobinson251:
Fwiw `any(hasmissing, vecs)` reads as morning understandable to me then `any(anymissing, vecs)`

nalimilan:
What is `vecs`?

nickrobinson251:
> What is `vecs`?

any iterable of vectors, e.g. `eachcol(df)`. i.e.
`any(hasmissing, eachcol(df))` reads as more understandable to me then any(anymissing, `eachcol(df))`

nalimilan:
But we just said we would define `hasmissing(::DataFrame)`/`anymissing(::DataFrame)` for convenience...

nickrobinson251:
i was just voiceing a weak preference for `hasmissing` over `anymissing`. I was just trying to give an example of a case where it read better to me, hence using some other name `vecs`. `vecs` could be `Vector{<:Vector}` if you prefer. Anyway, i've voiced my mild preference.

arnavs:
I'm open to `anymissing`, but also weakly prefer `hasmissing`.

bkamins:
I am OK with `hasmissing` because it reads more cleanly, however, `anymissing` (or `containsmissing` - but this is verbose) is a more precise name in my opinion.

`hasmissing(x)` suggests that we check some "property" of `x` that `x` owns (has) is `missing`. Like `hasproperty`, `hasfield`, `hasmethod`.

While `missing` value is not something that `x` owns but something that is contained in `x`. So we ask "Does `x` contain any element that is `missing`?" not "Does `x` have `missing`?" (or in a positive sentence we say "`x` contains a `missing`" not "`x` has a `missing`")

arnavs:
Fair enough, I do see the argument for `anymissing`.

ararslan:
To me, `anymissing(x)` doesn't seem beneficial over `any(ismissing, x)`, both in terms of the name and the functionality.

arnavs:
@ararslan what about the dataframe example? Currently you'd need to do `any(x -> any(ismissing, x), eachcol(df)`

ararslan:
Yeah, I don't see anything wrong with that; I've written something nearly identical within the last few weeks in a different context (nested JSON dicts).

arnavs:
There's nothing wrong with it, necessarily. But we're getting into regions where the syntax is seriously clunky.

Anyways, will let you guys decide.

mschauer:
In my view, `any(ismissing, x)` is a perfectly fine idiom, but a bit of a didactical problem. I can see how this is not the first thing a user will think of when running into the issue. How can we address that?

mschauer:
I know understand why some people were unhappy with `hasmissing` and agree. I liked it immediately because it sounds so familiar, but it is familiar for the wrong reasons: 
With @nalimilan, `has` has prefix signals a property of argument not its contents. Compare `HasEltype`, `HasLength`,  `HasShape`, `hasfield`, etc. with the exception `haskey`.

arnavs:
Yeah, I was instinctively repelled by `anymissing`, but I see the value of it. It's not as unnatural as I thought. 

I will move this PR over to `anymissing` and then hopefully @nalimilan or others can merge.

arnavs:
(Actually, I believe @bkamins or someone could also merge, since @nalimilan indicated he's fine with the PR except for the name change, which has been done.)

bkamins:
@arnavs - I understand that you want to have this PR merged quickly (I know the pain - I currently have 12 PRs that are waiting for a review and a possible merge) and thank you for involvement in the development. However, please note that:

1. what we do here will likely stay for years and be used by thousands of users so usually there is a period during which the PR (especially like this one - that modifies the API; a different story is bug-fixing PRs which have a quicker turnover usually) is kept open to let people think about it and comment (e.g. not all people involved in the development of the ecosystem and might want to comment have time to work on open source projects every day).
2. even if the PR is merged it will be only available on master; there will be another step (and an associated decision) of making a release of the package before it officially goes public.

From my experience - this PR has already received a lot of attention of the community, so I do not think that there is a risk that it will get forgotten and stay unmerged if it is eventually positively evaluated.

arnavs:
@bkamins Thanks for the reply. Respectfully, I still don't see that anything in my original comment is untrue (you wanted to persuade someone; they're now persuaded.) 

My only other experience with submitting usability PRs to Julia (`eachcol`, `eachrow`, etc.) descended into a morass of bikeshedding, which was mildly frustrating to most people involved and which I'd rather not see repeated. 

But sure, if you feel it is helpful to wait even further, I'm happy to do that.

bkamins:
> what more is required?

Another approval from a member of JuliaData (and that member will probably simply merge the PR as there is already my approval on it).

arnavs:
Fair enough, makes sense. Thanks for all your help

andreasnoack:
FWIW I agree with https://github.com/JuliaData/Missings.jl/pull/114/#issuecomment-606207826. I don't think it's worth adding a new function for this when the current syntax is so simple.

nalimilan:
Agreed. I've also realized that we cannot define `anymissing(::DataFrame)` as data frames are treated as collections of rows, not of cells -- and a row is never `missing`.

arnavs:
I'll close this then. I still agree with @nickrobinson251's point above: 

> I think this is a nice little addition, that's faster in some otherwise very slow cases, and simplifies a common patern for some users 

And think that the same argument against this function (why not use `any(ismissing, a)`?) also holds for `ismissing` (why not use `isa(x, Missing)`?) 

But if DataFrames are out, then the key feature I had in mind is out too. And the PR is out of date anyway. So may as well close.

nalimilan:
Regarding performance, if you find cases which are slower than they should, please file an issue in Base so that we add specializations for `any(::typeof(ismissing), x)`.

PR #119, opened 2020-08-24
Update version from `0.4.3` to `0.5.0`

pdeffebach:
I think adding a new function feature like `skipmissings` qualifies as a `0.4` to `0.5` jump according to SemVer since code written for `0.4.4` is expected to work with `0.4.3`.


nalimilan:
Actually, before 1.0, bumping the minor version number is considered as a breaking version, so we should only update the patch version.

nalimilan:
https://github.com/JuliaRegistries/General/pull/20267

Issue #121, opened 2020-10-09
Add a function for skippmissing and then "spread"ing the result


pdeffebach:
This is something I have discussed elsewhere as being useful, and came up today on slack. 

Problem

```
function standardize(x)
    mu = mean(x)
    sigma = std(x)
    return (x .- mu) ./ sigma
end
standardize([1,2,3, missing]
```

We can't use `passmissing` because this isn't an element-wise operation. We need an operation that applies skipmissing to the input, then applies the function, then "spreads" the result of the function to a vector of the same length as the output. 

```
function skipmissing_then_collect(fun, args...)
    smargs = skipmissings(args...)
    res = fun(smargs...)
    out = Union{eltype(res), Missing}[missing for i in 1:length(first(args))]# assume all args vectors also
    res_counter = 1
    for i in eachindex(first(smargs))
        out[i] = res[res_counter] # can probably do fancy iteration stuff here
        res_counter += 1
    end
    out
end
```

This might solve a lot of problems in DataFrames as well.

bkamins:
Great - let us discuss it here. This would be much better to have a generic solution (if we can come up with a good recipe) than having a custom patch in DataFrames.jl.

nalimilan:
As noted when discussing this in DataFrames (https://github.com/JuliaData/DataFrames.jl/issues/2258#issuecomment-655096412), it could be better (when inputs are vectors) to first find the indices of complete observations, pass a `SubArray` view of the complete observations to the user-provided function, and use these indices to fill the returned vector. That would avoid going over the data twice to identify missing values, and that would be simpler for users since the function would be passed `AbstractVector`s rather than `SkipMissings` iterators.

PR #122, opened 2020-10-31
Add `spreadmissings`

pdeffebach:
Adds behavior described in #121 

Implementation is easy enough to follow

```
function (f::SpreadMissings{F})(xs...) where {F}
    @assert all(x -> x isa AbstractVector, xs)

    x1 = first(xs)
    @assert all(x -> length(x) == length(x1), xs)

    if any(x -> x isa AbstractVector{>:Missing}, xs)
        nonmissinginds = collect(eachindex(skipmissings(xs...)))
        # TODO: narrow the type of the `view` to exclude `Missing`
        res = f.f((view(x, nonmissinginds) for x in xs)...)

        out = Vector{Union{Missing, eltype(res)}}(missing, length(x1))
        out[nonmissinginds] .= res

        return out
    else
        return f.f(xs...)
    end
end
```

### DIFF
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -2,7 +2,7 @@ module Missings
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
        Missing, MissingException, levels, coalesce, passmissing, nonmissingtype,
-       skipmissings
+       skipmissings, spreadmissings
 
 using Base: ismissing, missing, Missing, MissingException
 using Base: @deprecate
@@ -208,6 +208,80 @@ missing
 """
 passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
+struct SpreadMissings{F} <: Function
+    f::F
+end
+
+function (f::SpreadMissings{F})(x) where {F}
+    @assert x isa AbstractVector
+
+    if x isa AbstractVector{>:Missing}
+        nonmissinginds = collect(eachindex(skipmissing(x)))
+        res = f.f(view(x, nonmissinginds))
+
+        out = Vector{Union{Missing, eltype(res)}}(missing, length(x))
+        out[nonmissinginds] .= res
+
+        return out
+    else
+        return f.f(x)
+    end
+end
+
+function (f::SpreadMissings{F})(xs...) where {F}
+    @assert all(x -> x isa AbstractVector, xs)
+
+    x1 = first(xs)
+    @assert all(x -> length(x) == length(x1), xs)
+
+    if any(x -> x isa AbstractVector{>:Missing}, xs)
+        nonmissinginds = collect(eachindex(skipmissings(xs...)))
+        # TODO: narrow the type of the `view` to exclude `Missing`
+        res = f.f((view(x, nonmissinginds) for x in xs)...)
+
+        out = Vector{Union{Missing, eltype(res)}}(missing, length(x1))
+        out[nonmissinginds] .= res
+
+        return out
+    else
+        return f.f(xs...)
+    end
+end
+
+"""
+    spreadmissings(f)
+
+Return a function that performs the following transformation given the input `args...`
+
+1. Assumes all `args` are `AbstractArray`s
+2. Determines the indicies of `args` for which all elements are non-missing
+3. Applies `f` on the `view`s or `args` for the non-missing indices, returning `res`
+4. Returns a result the same length as the elements in `args`, filling in `missing`
+   as needed
+
+# Examples
+```
+julia> x = [0, 1, 2, missing]; y = [-1, 0, missing, 2];
+
+julia> function restricted_fun(x, y)           map(x, y) do xi, yi
+               if xi < 1 || yi < 1 # will error on missings
+                   return 1
+               else
+                   return 2
+               end
+           end
+       end;
+
+julia> spreadmissings(restricted_fun)(x, y)
+4-element Vector{Union{Missing, Int64}}:
+ 1
+ 1
+  missing
+  missing
+```
+"""
+spreadmissings(f) = SpreadMissings{Core.Typeof(f)}(f)
+
 """
    skipmissings(args...)
 
@@ -258,7 +332,7 @@ struct SkipMissings{V, T}
     others::T
 end
 
-Base.@propagate_inbounds function _anymissingindex(others::Tuple{Vararg{AbstractArray}}, i)    
+Base.@propagate_inbounds function _anymissingindex(others::Tuple{Vararg{AbstractArray}}, i)
    for oth in others
         oth[i] === missing && return true
     end
@@ -267,7 +341,7 @@ Base.@propagate_inbounds function _anymissingindex(others::Tuple{Vararg{Abstract
 end
 
 @inline function _anymissingiterate(others::Tuple, state)
-    for oth in others 
+    for oth in others
         y = iterate(oth, state)
         y !== nothing && first(y) === missing && return true
     end
@@ -278,7 +352,7 @@ end
 const SkipMissingsofArrays = SkipMissings{V, T} where
     {V <: AbstractArray, T <: Tuple{Vararg{AbstractArray}}}
 
-function Base.show(io::IO, mime::MIME"text/plain", itr::SkipMissings{V}) where V 
+function Base.show(io::IO, mime::MIME"text/plain", itr::SkipMissings{V}) where V
     print(io, SkipMissings, '{', V, '}', '(', itr.x, ')', " comprised of " *
           "$(length(itr.others) + 1) iterators")
 end
@@ -335,7 +409,7 @@ end
 @inline function Base.getindex(itr::SkipMissingsofArrays, i)
     @boundscheck checkbounds(itr.x, i)
     @inbounds xi = itr.x[i]
-    if xi === missing || @inbounds _anymissingindex(itr.others, i) 
+    if xi === missing || @inbounds _anymissingindex(itr.others, i)
         throw(MissingException("the value at index $i is missing for some element"))
     end
     return xi
@@ -380,9 +454,9 @@ Base.mapreduce_impl(f, op, A::SkipMissingsofArrays, ifirst::Integer, ilast::Inte
     A = itr.x
     if ifirst == ilast
         @inbounds a1 = A[ifirst]
-        if a1 === missing 
+        if a1 === missing
             return nothing
-        elseif _anymissingindex(itr.others, ifirst) 
+        elseif _anymissingindex(itr.others, ifirst)
             return nothing
         else
             return Some(Base.mapreduce_first(f, op, a1))
@@ -436,7 +510,7 @@ end
 Return a vector similar to the array wrapped by the given `SkipMissings` iterator
 but skipping all elements with a `missing` value in one of the iterators passed
 to `skipmissing` and elements for which `f` returns `false`. This method
-only applies when all iterators passed to `skipmissings` are arrays. 
+only applies when all iterators passed to `skipmissings` are arrays.
 
 # Examples
 ```


pdeffebach:
I just did some re-factoring. New implementation is

```
function (f::SpreadMissings{F})(xs...) where {F}
    if any(x -> x isa AbstractVector{>:Missing}, xs)
        vecs = Base.filter(x -> x isa AbstractVector, xs)
        s = skipmissings(vecs...)
        vecs_counter = 1
        newargs = ntuple(length(xs)) do i
            if xs[i] isa AbstractVector
                t = s[vecs_counter]
                vecs_counter += 1
            else
                t = xs[i]
            end
            t
        end

        res = f.f(newargs...)

        if res isa AbstractVector
            out = Vector{Union{Missing, eltype(res)}}(missing, length(first(vecs)))
            out[collect(eachindex(first(s)))] .= res
        else
            out = Vector{Union{Missing, typeof(res)}}(missing, length(first(vecs)))
            out[collect(eachindex(first(s)))] .= Ref(res)
        end

        return out
    else
        return f.f(xs...)
    end
end
```

Miraculously, this infers function infers.

pdeffebach:
Okay after playing around with this here's what I think an API should be. Consider the call 

```
f(v1::Vector, v2::Vector, s::Scalar)
``` 

Recall that `skipmissings(x, y)` returns a `tuple` of iterators `sx, sy` that iterate through `x` and `y` where both indices are not `missing`. We need


1. `skipfun(f)` such that `skipfun(f)(v1, v2, s)` is 

```
sv1, sv2 = skipmissings(v1, v2)
f(sv1, sv2, x)
```

It automatically finds the vector values and calls `skipmissings` on those. Maybe we can dig into the broadcasting infrastructure to make this work more intelligently?

2. `spreadmissings`, defined in this PR. `spreadmisings(f)(v1, v2, s)` is equivalent to 

```
sv1, sv2 = skipmissings(v1, v2)
res = f(sv1, sv2, x)
if res isa AbstractVector
    out = Vector{Union{Missing, eltype(res)}}(missing, length(v1))
    out[collect(eachindex(sv1))] .= res
else
    out = Vector{Union{Missing, typeof(res)}}(missing, length(v1))
    out[collect(eachindex(sv1))] .= Ref(res)
end
out
```


The problem with `spreadmissings` is that it is restricted to Vectors and it's hard to figure out all the types that really need to be "spread". Consider, for example 

```
function maketable(v1, v2, s)
    # stuff which errors on missing values
    (n1 = v1 .+ v2 .* s, n2 = v1 .- v2 .* s)
end
```

If we use `spreadmissings` logic, we know that this should be a named tuple of vectors each of length `v1`. Unfortunately we can't just special case a named tuple of vectors to be the kind of thing that "spreads". I mean, maybe we can, but it's easy to see how that kind of logic can get more and more complicated.

pdeffebach:
We also need

1. `passmissingfalse` which returns `false` instead of `missing`, for use in filtering and `&&` etc. 
2. `spreadmissingfalse` which fills in `false` instead of `missing` in the missing indices.

matthieugomez:
Maybe it would be more elegant to restrict it to functions for which all arguments are AbstractVector, i.e. `@assert all(x -> x isa AbstractVector, xs)`.
AFAIU, this does not hurt usability in DataFrames since users need to create anonymous function with columns anyway. Same thing for DataFramesMeta.

pdeffebach:
Thats what I had initially. Then I thought about broadcasting and how it automatically figures out what should be broadcasted. So I updated this PR to emulate that behavior a little.

matthieugomez:
Another thing is that `spreadmissing` always returns a `Vector`, but som  function may return `AbstractVectors` that are not `Vectors` (e.g. . `cut` returns a `CategoricalVector`) 

So it may be better to use in `spreadmissing` a function like `extend_missings` which extends an AbstractVector to accept  missings, i.e.
```julia
function extend_missings(v::AbstractVector{T}, esample::BitVector)
   out = Vector{Union{Missing, T}(missing, length(esample))
   out[esample] = v
   return out
end
```
The point is that packages can defined specialized methods of `extend_missings` for their particular type.

pdeffebach:
That's a good point. Maybe I will try to formalize all of this behavior into a separate package first and see how it goes.

matthieugomez:
In any case, it'd be great to define such a function! It's just hard to find the right abstraction, so it's good to experiment.

src/Missings.jl
```diff
@@ -208,6 +208,104 @@ missing
...
+        return out
+    else
+        return f.f(x)
+    end
+end
+
+function (f::SpreadMissings{F})(xs...) where {F}
+    if any(x -> x isa AbstractVector{>:Missing}, xs)
+        vecs = Base.filter(x -> x isa AbstractVector, xs)
+        s = skipmissings(vecs...)
```
nalimilan, line 234: As we discussed somewhere in a DataFrames issue, I don't think using `skipmissings` here is really efficient: it's going to require two passes over the data: one during the call to `f`, and one for `collect(eachindex(first(s)))`. Since you do the latter step, better use the result to create `SubArray` views of arguments, which are faster than `skipmissings` and more convenient for users.

src/Missings.jl
```diff
@@ -208,6 +208,104 @@ missing
...
+        out = Vector{Union{Missing, eltype(res)}}(missing, length(x))
+        out[nonmissinginds] .= res
+
+        return out
+    else
+        return f.f(x)
+    end
+end
+
+function (f::SpreadMissings{F})(xs...) where {F}
```
nalimilan, line 231: Also support keyword arguments?

src/Missings.jl
```diff
@@ -208,6 +208,104 @@ missing
 """
 passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
+struct SpreadMissings{F} <: Function
+    f::F
+end
+
+function (f::SpreadMissings{F})(x) where {F}
```
nalimilan, line 215: Do you really need a specific method for a single `x`?

src/Missings.jl
```diff
@@ -208,6 +208,104 @@ missing
...
+
+        return out
+    else
+        return f.f(x)
+    end
+end
+
+function (f::SpreadMissings{F})(xs...) where {F}
+    if any(x -> x isa AbstractVector{>:Missing}, xs)
+        vecs = Base.filter(x -> x isa AbstractVector, xs)
```
nalimilan, line 233:
```suggestion
        vecs = filter(x -> x isa AbstractVector, xs)
```

nalimilan:
Requiring packages to extend `extend_missings` won't really work as they would all need to add a dependency on Missings, which most won't want to do. Why not just call `allowmissing(res)`, which uses `similar` and should therefore return the right array type? for the case where a scalar is returned, `similar(vecs[1], Union{Missing, typeof(res)})` shouldn't be too bad (at least it will return a `CategroicalArray` if `res isa CategoricalValue`).

src/Missings.jl
pdeffebach: no, its unnecessary.

src/Missings.jl
pdeffebach: Seems needed for some reason, so not changed.

pdeffebach:
> which uses `similar` and should therefore return the right array type? for the case where a scalar is returned, `similar(vecs[1], Union{Missing, typeof(res)})` shouldn't be too bad (at least it will return a `CategroicalArray` if `res isa CategoricalValue`).

I changed it, but I don't think the implementation is quite correct. In particular, I'm worried about the scalar case. I'm not sure what the importance of the first argument in `similar(x, eltype, length)` is. What's the purpose of `vecs[1]` in `similar(vec[1], Union{typeof(res), Missing})`?

pdeffebach:
I think it would be totally awesome if @nalimilan or @quinnj took over this PR. It's over my head to deal with all the performance considerations and necessary benchmarking. 

If possible maybe we can merge it as an undocumented feature and see if it works?

bkamins:
@nalimilan - please ping me when the first round of comments is resolved. Then I will review. Thank you!

src/Missings.jl
```diff
@@ -208,6 +208,91 @@ missing
...
+struct SpreadMissings{F} <: Function
+    f::F
+end
+
+function (f::SpreadMissings{F})(xs...; kwargs...) where {F}
+    if any(x -> x isa AbstractVector{>:Missing}, xs)
+        vecs = Base.filter(x -> x isa AbstractVector, xs)
+        s = skipmissings(vecs...)
+        nonmissinginds = collect(eachindex(first(s)))
+        vecs_counter = 1
```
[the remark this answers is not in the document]
pdeffebach: It's used below, to skip non-vector arguments.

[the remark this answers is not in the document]
nalimilan: Sorry, I don't see it. All I can see is `vecs_counter += 1`.

src/Missings.jl
```diff
@@ -219,17 +219,15 @@ function (f::SpreadMissings{F})(xs...; kwargs...) where {F}
         findex = eachindex(first(vecs))
         @assert all(x -> eachindex(x) == findex, vecs)
 
-
         nonmissingmask = fill(true, length(vecs[1]))
         for v in vecs
-            nonmissingmask .&= .!ismissing(v)
+            nonmissingmask .&= .!ismissing.(v)
         end
-        nonmissinginds = findall(nonmissingmask)
```
nalimilan: Why remove this? In general I think it's faster to compute indices manually when there are several columns, as otherwise indexing will have to compute the number of `true` values in the mask once for each indexing operation. Probably worth benchmarking.

src/Missings.jl
```diff
@@ -215,11 +215,17 @@ end
...
     if any(x -> x isa AbstractVector{>:Missing}, xs)
         vecs = Base.filter(x -> x isa AbstractVector, xs)
-    nonmissingmask = fill(true, length(vecs[1]))
-    for v in vecs
-        nonmissingmask .&= .!ismissing(v)
-    end
-    nonmissinginds = findall(nonmissingmask)
+
+        findex = eachindex(first(vecs))
+        @assert all(x -> eachindex(x) == findex, vecs)
```
nalimilan, line 220: Can you throw a more explicit error in case of failure? It would be nice to mention the indices which do not match.

pdeffebach:
Okay after the merging of `@rtransform` etc. in DataFramesMeta, this will be my next focus. 

just to confirm, do we want this new feature to live in Missings.jl for now? Or do we want me to move all this stuff to DataFramesMeta.

bkamins:
Can you please summarize briefly the consensus how things should work? Thank you!

nalimilan:
I think it would be good to have `spreadmissings` as an internal function in DataFramesMeta for a while until we are sure the API is what we need. Development can continue in this PR so that it can be merged later.

pdeffebach (commented):
> Can you please summarize briefly the consensus how things should work? Thank you!

Okay, the plan for now is to have a single `@passmissing` flag in DataFramesMeta. 

When `@byrow` is used, we get `ByRow(passmissing(f))`. When we are operating by columns, we use this function, which takes `view`s of the indices in the columns where all values are non-missing.

pdeffebach (commented):
Still have more work to do.

Manifest.toml
```diff
@@ -0,0 +1,9 @@
+# This file is machine-generated - editing it directly is not advised
```
nalimilan, line 1: Remove this.

src/Missings.jl
```diff
@@ -213,11 +213,32 @@ struct SpreadMissings{F} <: Function
 end
 
 function (f::SpreadMissings{F})(xs...; kwargs...) where {F}
+
     if any(x -> x isa AbstractVector{>:Missing}, xs)
         vecs = Base.filter(x -> x isa AbstractVector, xs)
 
         findex = eachindex(first(vecs))
-        @assert all(x -> eachindex(x) == findex, vecs)
+        if !(all(x -> eachindex(x) == findex, vecs[2:end]))
```
nalimilan, line 222:
```suggestion
        if !(all(x -> eachindex(x) == findex, view(vecs, 2:end)))
```

src/Missings.jl
```diff
@@ -208,6 +208,91 @@ missing
 """
 passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
+struct SpreadMissings{F} <: Function
+    f::F
+end
+
+function (f::SpreadMissings{F})(xs...; kwargs...) where {F}
+    if any(x -> x isa AbstractVector{>:Missing}, xs)
+        vecs = Base.filter(x -> x isa AbstractVector, xs)
```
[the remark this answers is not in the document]
nalimilan: Yeah. But indeed that would be quite restrictive. Maybe we can hardcode a list of allowed types for now, or conversely throw an error for some collection types like `Tuple`, etc.? It's not clear to me yet whether it would be more useful to treat them like scalars or like vectors.

The list of hardcoded behaviors for `broadcasted` is here:
https://github.com/JuliaLang/julia/blob/8197c418e1d806f5faf1fd900c78068ea8046e1d/base/broadcast.jl#L700-L705

src/Missings.jl
```diff
@@ -208,6 +208,91 @@ missing
...
+        else
+            out = similar(vecs[1], Union{typeof(res), Missing})
+            fill!(out, missing)
+            out[nonmissinginds] .= Ref(res)
+        end
+
+        return out
+    else
+        return f.f(xs...; kwargs...)
+    end
```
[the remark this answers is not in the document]
nalimilan: OK. Something that occurred to me though: should we do something special if one of the non-vector arguments is `missing` (like returning `missing`)? Should we reserve this behavior for the future in case we want `spreadmissings` to be a more general version of `passmissing`?

src/Missings.jl
```diff
@@ -208,6 +208,128 @@ missing
...
+                ind_msg = "Vector inputs $inds have indices $k \n"
+                s = s * ind_msg
+            end
+
+            throw(ArgumentError(s))
+        end
+
+        nonmissingmask = fill(true, length(vecs[1]))
+        for v in vecs
+            nonmissingmask .&= .!ismissing.(v)
```
nalimilan, line 246: As I noted in an outdated comment, it would be faster to do `nonmissinginds = findall(nonmissingmask)` and use that below instead of `nonmissingmask`, which has to recompute integer indices each time it's used.

src/Missings.jl
```diff
@@ -208,6 +208,128 @@ missing
...
+            for i in 1:length(vecs)
+                e = eachindex(vecs[i])
+                if eachindex(e) in keys(d)
+                    push!(d[i], i)
+                else
+                    d[e] = [i]
+                end
+            end
+            s = "The indices of vector-input arguments are not all " *
+                "the same.\n"
```
nalimilan, line 233:
```suggestion
            s = "The indices of vector input arguments are not all the same.\n"
```

src/Missings.jl
```diff
@@ -208,6 +208,128 @@ missing
...
+
+            for k in keys(d)
+                inds = join(d[k], ", ", " and ")
+                ind_msg = "Vector inputs $inds have indices $k \n"
```
nalimilan, line 237:
```suggestion
                ind_msg = "Vector inputs $inds have indices $k\n"
```

src/Missings.jl
pdeffebach: Added.

src/Missings.jl
pdeffebach: `vecs` is a small tuple, here. So view doesn;t work.

src/Missings.jl
pdeffebach: Added.

src/Missings.jl
pdeffebach: Added.

src/Missings.jl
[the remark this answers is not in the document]
pdeffebach: I don't think so. We definitely can't return `missing`, we would have to return a vector of `missing`s, which seems very restrictive.

src/Missings.jl
```diff
@@ -208,6 +208,91 @@ missing
...
+```
+sx, sy = skipmissings(x, y)
+```
+
+and call
+
+```
+f(sx, y, sy)
+
+```
```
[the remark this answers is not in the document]
pdeffebach: Changed the text.

src/Missings.jl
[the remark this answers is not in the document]
pdeffebach: Okay let me know if you think the updated `non_spreadable_check` is okay.

pdeffebach (commented):
Thanks. Please confirm I addressed this round of concerns.

nalimilan:
I think you forgot pushing. :-)

src/Missings.jl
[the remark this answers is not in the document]
nalimilan: Not sure what "restrictive" means here. That would be similar to e.g. `broadcast(+, missing, [1, 2, 3])`.

Whether that makes sense depends on what `f` does with its inputs, but I can't find examples of functions to which one would pass one or more vectors plus a `missing` argument and would not expect the resulting vector to be full of `missing`s. So reserving the situation where one argument is `missing` could be a good idea in case we want to handle it later.

pdeffebach:
I think I `push`ed, then made comments after, so it looks like I forgot to push but I did. At least, the current state of the PR matches what's on my local computer. 

Would love someone to take this over though.

bkamins:
I will have a look at it in April and then try to discuss with @nalimilan how to proceed. Thank you!

bkamins:
I went through the PR and thought a bit about it. I would propose - to make sure we can make some constructive progress on it:
1. can we list functions which we think `spreadmissings` would be used with?
2. My intuition is that kwargs should be left untouched (they are usually passed as supplementary information to functions, eg. in plotting you can get a vector there that you do not want to be touched)
3. for now I would assume that all arguments must be `AbstractVector`. I know it is restrictive, but this makes sure that anything we add in the future to this function will be non-breaking. Also it seems for me (see point 1. and some discussion above) that this would be the most common use-case.

pdeffebach:
> 1\. can we list functions which we think `spreadmissings` would be used with?

I think this function is well used with Statistical functions, like `cor` and `cov`. I know @nalimilan put a lot of effort into `pairwise`, but there are still some gaps, in particular using multiple arrays and getting `cov` to return a scalar rather than a matrix. 

It is also useful for plotting functions. There was someone on Discourse yesterday who couldn't use UnicodePlots.jl because their vectors had `missing` values in them. 


> 2\. My intuition is that kwargs should be left untouched (they are usually passed as supplementary information to functions, eg. in plotting you can get a vector there that you do not want to be touched)

I disagree. For example, a `weights` keyword argument will need to be synced with the positional arguments. Besides, users can always use `Ref` to make them ignored. 

> 3\. for now I would assume that all arguments must be `AbstractVector`. I know it is restrictive, but this makes sure that anything we add in the future to this function will be non-breaking. Also it seems for me (see point 1. and some discussion above) that this would be the most common use-case.

Maybe. I would say I'm okay with an initial version that left keyword arguments untouched and required vectors, but that would induce a breaking change if we were to modify keyword arguments too. I'm inclined to use implement current behavior and mark it as experimental. 

Another issue is the "spread"-ing of the result. Currently, we *always* return an array, which would not work for `cor` and for `plotting`. I propose a keyword argument to `spreadmissing(f; :spread = ...)` where `spread` is 

* `:vector`: Spread out missing values if the returned object is an `AbstractVector` and don't spread otherwise
* `:none`, *never* spread, this is useful if you have, say `model_fit(x, y)` which returns a Parameter vector which is not the same size as `x` and `y`. 
* `:all:` Spread if there is a vector, and for any non-vector outputs, put them into a `vector` with `missing`s filled in.

bkamins:
I have re-read the docstring:

> will find the indices which corresond to `missing` values in *both*
> `x` and `z`. Then apply `f` on the `view`s of `x` and `z` which
> contain non-missing values.

Given this I think we can only accept `AbstractVector` for skipping missing values. The reason is that you require the collection to be lineary indexable with a `view` and this is essentially `AbstractVector`.

Second is protection from spreading missing values - `Ref` is mentioned in your comment, but I think we need a different type for this, as user might pass `Ref` value and we could not distinguish it.

Finally I do not fully understand the `spread` argument of `spreadmissing`. I thought that we did not want `spreadmissing` to affect the output of the passed function (or am I missing something?). Do you mean that if I do `spreadmissing(+)` and you pass `:vector` then doing `spreadmissing(+, spread=:vector)([1, 2, missing], [3, missing, 4])` would produce `[4, missing, missing]` and with `:none` would produce just `[4]`?

pdeffebach:
> Given this I think we can only accept `AbstractVector` for skipping missing values. The reason is that you require the collection to be lineary indexable with a `view` and this is essentially `AbstractVector`.

Yes, we only apply the skipping to `vectors`. But other arguments are left untouched. Let's say there is a function `normalize` where the author unfortunately used too-strict a type signature

```
julia> function normalize(
           m::Int, 
           x1::AbstractVector{<:Real}, 
           x2::AbstractVector{<:Real})
           
           min.(m, x1, x2)
       end;
```

we should be able to do

```
julia> x = [1, 2, missing, 4]; y = [5, 6, 7, missing];

julia> spreadmissings(normalize)(1, x, y)
```

Note: Currently this doesn't work :( because we don't change the `eltype` of the view... Maybe @nalimilan knows if there is anything we can do to get around this. 

> Second is protection from spreading missing values - `Ref` is mentioned in your comment, but I think we need a different type for this, as user might pass `Ref` value and we could not distinguish it.

Okay. Not sure the best way to do this, then. defining a new protection type seems like overkill. Is there a standard way to do this?

> Finally I do not fully understand the `spread` argument of `spreadmissing`. I thought that we did not want `spreadmissing` to affect the output of the passed function (or am I missing something?). Do you mean that if I do `spreadmissing(+)` and you pass `:vector` then doing `spreadmissing(+, spread=:vector)([1, 2, missing], [3, missing, 4])` would produce `[4, missing, missing]` and with `:none` would produce just `[4]`?

Yes, this is the behavior we want. It's really useful in tabular applications. Think of the `normalize` function above. If we didn't allow `spread`ing, then you wouldn't be able to use it to work with data frames or any other table.

bkamins:
> if there is anything we can do to get around this.

You would have to create your custom view type. This is what `skipmissing` does.

bkamins:
> Is there a standard way to do this?

Standard way is to create a new type. The alternative would be to use a macro instead of a function and then we can protect with whatever we like (probably best to be consistent with DataFramesMeta.jl).

pdeffebach:
> > if there is anything we can do to get around this.
> 
> You would have to create your custom view type. This is what `skipmissing` does.

I think there should be a way to circumvent the constructor of a `SubArray` to make the `eltype` do what we want. I will ask Slack.

bkamins:
> I think there should be a way to circumvent the constructor of a SubArray to make the eltype do what we want.

Indeed maybe there would be a workaround. But then you need to decide if the resulting type should have continuous indexing, or like with `skipmissing` discontinuous indexing.

pdeffebach:
I want to use this function internally inside `spreadmissings` (or as something exposed to users if we think it's useful)

```
julia> using Missings

julia> function unsafe_nonmissingview(x::Vector, inds...)
           SubArray{nonmissingtype(eltype(x)), 1, typeof(x), typeof(inds), false}(x, inds, 0, 0)
       end;

julia> x = [100, 200, missing, 400];

julia> foo(x::AbstractVector{<:Real}) = 1;

julia> t = unsafe_nonmissingview(x, [1, 2, 4]);

julia> foo(t)
1
```

It's unsafe because you have to *know* that your indices don't have any missing values. It would be great to get a Base dev's opinion on this.

bkamins:
I think that for internal purposes it is OK.

bkamins:
@aplavin - nice. How would it support multiple vectors (also - if I understand this correctly @pdeffebach wants a function wrapper like `passmissing`)

bkamins:
> I don't really understand what semantics are you looking for multiple vectors. Any examples?

We want `spreadmissings(cor)([1., missing, 2, 3, 4], [1., 2, missing, 3, 4])` to compute `cor([1., 3, 4], [1., 3, 4])` (there are also other cases but this is the major one @pdeffebach wanted to cover).

bkamins:
> how does it "spread" anything?

It spreads `missing` across input vectors.

> `cor` returns a single number, 

That is why I have written "there are also other cases". @pdeffebach proposed above also:
```
spreadmissing(fun, spread=:vector)([1, 2, missing], [3, missing, 4])
```
However, I am not sure that this is that useful, as it can be just done currently using `passmissing` and broadcasting. But maybe there are cases when it is useful.

bkamins:
There were some issues with `skipmissings` design (I do not remember exactly what). Probably when @pdeffebach comments he can present a complete state of this issue.

pdeffebach:
Here are the issues this is trying to solve (and this PR might be going the wrong direction in solving all of them at once)

1. We want `cor([1, 2, missing, 4], [5, 6, 7, missing])` to compute `cor([1, 2], [5, 6])`. That is, we want to match indices of non-missing values. 
2. Currently, we can't rely on `::Any` iterators like `skipmissings` due to dispatch issues with `cor`:

```
julia> using Missings, Statistics;

julia> x = [1, 2, missing, 4]; y = [5, 6, 7, missing];

julia> sx, sy = skipmissings(x, y);

julia> cor(sx, sy)
ERROR: MethodError: no method matching cor(::Missings.SkipMissings{Vector{Union{Missing, Int64}}, Tuple{Vector{Union{Missing, Int64}}}}, ::Missings.SkipMissings{Vector{Union{Missing, Int64}}, Tuple{Vector{Union{Missing, Int64}}}})
```

Once option would be to add support for any iterator. But Base devs are not supportive of that going in Statistics.jl. See [here](https://github.com/JuliaStats/Statistics.jl/pull/34#issuecomment-622062726). 

More generally, we want users who have vectors of missing values to be able to work with APIs that have overly restrictive type signatures (`AbstractVector{<:Real}`), like some `histogram` in Plots.jl (or other examples, if that particular one is fixes). 

So we need to make sure we pass `View`s or similar to the underlying function. 

3. When working with Tabular data, we need to make sure operations that are `R^N -> R^N` work with this functionality, hence the "spread"-ing of the output. If the function returned the length of the non-missing array, then we couldn't use it with any tabular transformations. It looks like `@modify` from Accessors.jl solves this problem, but doesn't solve the two points above.

pdeffebach:
Yeah that's basically it. 

I think the re-assembly is important, though, so this PR should be split in 2

1. A `view`-based AbstractArray with missing indices. This is basically a super-charged `skipmissings`. 
2. This PR, which does both the skipping and the re-assembly.

nalimilan:
Supporting these two features with a single function indeed makes it a bit tricky, but I'm concerned that users may be confused by the need to use two different functions in situation that may seem to them really similar: e.g. `spreadmissings(normalize)(x, y)` vs. `passmissings(cor)(x, y)` (the latter would be a vectorized version of `passmissing`, which just returns the result without spreading/reassembling it). Having a single function which would "spread"/"reassemble" vector outputs, but return other types as they are would probably be more intuitive. A `spread` argument like @pdeffebach proposed above can allow choosing the behavior you want and ignore the result type for rare cases where it's needed.

Good point with passing a view with a non-`Missing` eltype. Though that can also be done later as an improvement.

pdeffebach:
taking a look at this. 

I need to add tests. I will get started on that even if the API isn't completely finalize. 

As for spreading missing values. My reading of the above comments is that we want three behaviors

* `:default`: If the output is a vector, create a new vector and fill in `missing` values as needed. Potentially do some check that the output vector has the same length as the number of non-missing values in all arguments. If the output doesn't match this criteria, error and suggest a different `spread` keyword argument. 

If the output is *not* a vector, don't spread the result and return as-is.

* `:always`: If the output is a vector, do the same behavior as above if the shapes match. If the shapes don't match, maybe return a `Vector` of `Vector`s? If the output is not a vector, make a `Vector` with `missing` values in the appropriate indices.

* `never`: Just return the output no matter what. 

This requires dispatching on a `Symbol`. Is that performant? Is there a better way to do this, maybe with `Val`s?

nalimilan:
Performance should be OK as the compiler will optimize the argument out if it's known at compile time, which is the case in general. That's what the `skipmissing` argument to `pairwise` does in StatsBase. To ensure this optimization works you can have a small `@inline` function which dispatches to separate functions depending on the value of the argument, or use different methods wrapping it in `Val`.

pdeffebach:
Okay implemented behavior and added tests. See the docstring below for the API

This is ready for a review @nalimilan @bkamins 

    spreadmissings(f; spread = :default)

Given a function `f`, function `f` but performs a transformation
on arguments to remove missing values before executing.

### Initial example

```julia-repl
julia> using Statistics;

julia> xmiss = [1, 2, 3, missing];

julia> ymiss = [missing, 200, 300, 400];

julia> summeans(x, y) = mean(x) + mean(y);

julia> spreadmissings(summeans)(xmiss, ymiss)
252.5
```

### Details

Given the call

```
spreadmissings(f)(x::AbstractVector, y::Integer, z::AbstractVector)
```

finds the indices which corresond to `missing` values in *both*
`x` and `z`. Then apply `f` on the `SubArray`s of `x` and `z` which
contain non-missing values. In essense:

```
inds = .!missing.(x) .& .!missing.(z)
sx = view(x, inds); sy = view(y, inds)
f(sx, y, sy)
```

!!! note
    `spreadmissings` does not use the default `view` behavior. Rather,
    it constructs a `SubArray` directly such that the eltype of the new
    inputs do not include `Missing`.

### `spread` keyword argument

Control over how the output from `f` is "spread"
along with respect to missing values.

* `:default`:
    * If `output` is a `Vector` with the same length as the number of
      jointly non-missing elements of the inputs `output` is "spread"
      to match the non-missing elements of the inputs.
    * If the `output` is a `Vector` whose length is not the same
      as the length of number of non-missing elements of the inputs,
      a `DimensionMismatch` error is thrown.
    * If the output is not a `Vector`, `output` is simply returned directly
* `:nonmissing`:
    * If `output` is a `Vector`, behavior is the same as `:default`
    * If `output` is not a `Vector`, `output` is spread along non-missing
    elements of the inputs.
* `:none`: `output` is returned directly, whether a `Vector` or not.

A summary of the behavior is given in the table below:

| spread \\ output type  | Vector           | Non-Vector       |
|:---------------------- |:---------------- |:---------------- |
| :default               | spread and match | return           |
| :nonmissing            | spread and match | spread and match |
| :none                  | return           | return           |

If there are `AbstractVector` inputs but none of these inputs
`AbstractVector{>:Missing}`, behavior of `spread` is the same as
with inputs which allows for missing values. However the returned
vectors will not allow for `missing`.

If none of the argumets are `AbstractVector`s, `spreadmissings(f)`
behaves the same as `f` regardpess of `spread`.

### Limitations

`spreadmissings` currently does not support:

* Different length vector inputs. For

```
spreadmissings(f)([1, 2], [100, 200, 300])
```

will error.

* Full spreading of scalar outputs across the *full* length of the
input vector. That is, there is no `spread = :all` option.

* No guarantees about the behavior of `CategoricalVector`s and other
specialized vector types. Though it does seem to work for now.

pdeffebach:
Can I get a review @nalimilan ?

jariji:
This function looks a bit complicated. Can we get minimal demo examples of all 6 cases?

pdeffebach:
Sure. 

1. Correlation between two vectors, each with missing values
```
julia> spreadmissings(cor)(x, y)
-0.026341269312980007
```

2. Apply a normalization across an entire vector, taking into account `missing`s.  This is feasible under current propagation rules. But let's add a type restriction so it would error otherwise. 
```
julia> normalize(x::AbstractVector{<:Real}) = x .- mean(x);

julia> spreadmissings(normalize)(x)[1:10]
10-element Vector{Union{Missing, Float64}}:
  0.35300498516309853
 -0.4455063763988256
  0.16127093602449238
   missing
  0.35862412095009555
  0.38201925729328823
  0.25052381186427586
 -0.38114014203916713
 -0.09463085885428646
  0.3900304330569576
```

3. "Spread" the `mean` of the output so that it only uses observations that contribute to the mean. 
```
julia> spreadmissings(mean; spread = :nonmissing)(x)[1:10]
10-element Vector{Union{Missing, Float64}}:
 0.518096315734334
 0.518096315734334
 0.518096315734334
  missing
 0.518096315734334
 0.518096315734334
 0.518096315734334
 0.518096315734334
 0.518096315734334
 0.518096315734334
```

4. Return a `Vector` but avoid automatic spreading of outputs. 
```
julia> extrema_vec(x) = collect(extrema(x));

julia> spreadmissings(extrema_vec; spread = :none)(x)
2-element Vector{Float64}:
 0.0005904666011299353
 0.999090318277402
```

jariji:
+1 for the default behavior. I wouldn't include the other two in this function, they're really separate concepts.

src/Missings.jl
```diff
@@ -208,6 +208,290 @@ missing
 """
 passmissing(f) = PassMissing{Core.Typeof(f)}(f)
 
+struct SpreadMissings{F} <: Function
+    f::F
+    spread::Symbol
+    function SpreadMissings(f, spread)
```
nalimilan, line 214:
```suggestion
    function SpreadMissings(f, spread::Symbol)
```

nalimilan (commented):
Sorry for not getting back to this earlier. Here are some comments. Looks mostly good!

A few final questions about the API:
- Calling this `spreadmissing` is weird for the case where nothing is spread. Calling it `passmissings` would be a logical alternative (extension of `passmissing` to vectors containing missing values), but it's also not super appropriate when the result isn't spread, as missing values are then skipped rather than passed. Maybe we could reclaim the name `skipmissings`, as it's been deprecated for some time?
- Regarding the `spread` keyword argument, maybe it would be clearer to have it take values `:vector` (or `:auto`), `:always` and `:never`, as you proposed at some point? Or it could take a type, with `AbstractVector`, `Any` and `Union{}`. 'm inclined to think that it would be simpler to have a Boolean argument with `true` (the default) indicating that vectors are spread, and `false` that they are not; non-vector results cannot be spread anyway so we don't really need three choices. Another solution would be to

CameronBieganek:
This function seems rather complicated. And I have doubts about its goals. Consider a function that centers a vector:

```julia
center(x) = x .- mean(x)
```

Here's what we get if we run it on a vector with `missing`:

```julia
julia> x = [1, 2, missing]
3-element Vector{Union{Missing, Int64}}:
 1
 2
  missing

julia> center(x)
3-element Vector{Missing}:
 missing
 missing
 missing
```

I would argue that that is the correct output. There's no way we can possibly center a vector if there are elements for which we do not know the value. In other words, if we do this,

```julia
julia> spreadmissings(center)(x)
3-element Vector{Union{Missing, Float64}}:
 -0.5
  0.5
   missing
```

then we are getting the *wrong* answer. Of course the user can decide that they want to drop `missing` and then center the vector, but they can do that with good old `skipmissing`:

```julia
julia> center(skipmissing(x))
2-element Vector{Float64}:
 -0.5
  0.5
```

The same argument applies to `zscore` and to the `normalize` function that was used as an example above.

So, I think it is better to focus on the multivariable functions case, like `cor`, `var`, and `summeans`. (Let's forget for the moment that `cor` and `var` only accept arrays.) The multivariable case can be covered pretty well with something like the following. (It looks pretty simple, but of course there is complexity hidden in the `skipmissings` implementation.)

```julia
function missingskipper(f)
    (args...; kwargs...) -> f(skipmissings(args...)...; kwargs...)
end
```

Now let's run it on `summeans`:

```julia
julia> summeans(x, y) = mean(x) + mean(y);

julia> x = [1, 2, 3, missing];

julia> y = [4, 5, missing, 6];

julia> missingskipper(summeans)(x, y)
6.0
```

I think it's cleaner to have `missingskipper(f)` apply `skipmissings` to every positional argument of `f`. That way we can allow arbitrary input types to `f`---we don't have to limit ourselves to only skipping missing values in `AbstractVector` inputs. If you need to ignore an argument, then use an anonymous function, like this:

```julia
missingskipper((x, z) -> f(x, 42, z))(x, z)
```

(This would be a great place for underscore currying.)

Initially, `missingskipper` might not be terribly useful for the average user, but hopefully it would be helpful for adding missing handling to the various dataframe macros. (I don't know for sure if this is true---I haven't really researched how `missingskipper` would be applied in DataFramesMeta.)

nalimilan:
The point is that we need the output to have the same number of elements as the input to store it as a data frame column. `df.xc = center(skipmissing(df.x))` doesn't work. `spreadmissing` is kind of a vectorized combination of `skipmissing` and `passmissing`. If that's not the behavior you want then of course don't use that function.

> There's no way we can possibly center a vector if there are elements for which we do not know the value.

I don't understand this point. If you take that stance, there's no way we can possibly compute the mean of the vector either. So basically using `skipmissing` would always be wrong, and `missingskipper` has the same problem.

> Initially, `missingskipper` might not be terribly useful for the average user, but hopefully it would be helpful for adding missing handling to the various dataframe macros. (I don't know for sure if this is true---I haven't really researched how `missingskipper` would be applied in DataFramesMeta.)

I don't see how your proposal would work in combination with data frame macros, as it would fail as soon as a non-collection argument is passed to a function. We precisely don't want people to have to write things like `missingskipper((x, z) -> f(x, 42, z))(x, z))`, which is IMO much harder to get than using another existing solution like `dropmissing`. And if the macro is supposed to handle this automatically, then `missingskipper` doesn't add anything compared to using `skipmissings`.

CameronBieganek:
> We precisely don't want people to have to write things like `missingskipper((x, z) -> f(x, 42, z))(x, z))`, which is IMO much harder to get than using another existing solution like `dropmissing`.

Limiting `spreadmissings` to only work on functions that take in `AbstractVector` is a deal breaker in my book. I want something more generic. As I mentioned in my first post, `missingskipper` is probably not very ergonomic for the casual user, but if we design it right then it might actually be useful to package developers.

CameronBieganek:
> More generally, we want users who have vectors of missing values to be able to work with APIs that have overly restrictive type signatures (AbstractVector{<:Real}), like some histogram in Plots.jl

I don't think it's a good idea to design an awkward API so that we can account for functions that have an awkward API. Better to encourage improved APIs in the ecosystem.

pdeffebach:
I heartily agree with Milan, here. 

1. `dropmissing(df)` isn't feasible when the user is working with a large data set and with many columns each of which has an irregular pattern of missing values. 
2. We can't have a solution which asks users to use a difficult syntax or file a github issue every time they encounter an error.  We need an easy syntax so users can analyze data and be productive. Plus, I'm sure there are lots of instances where package authors justifiably want to work with a `Vector` or want to avoid thinking about `missing`s. We provide a bridge between the data analyst and those packages. 
3. Putting it all in DataFramesMeta.jl is not a good outcome. Obviously I wrote this API with DataFramesMeta.jl in mind, but the core benefit of DataFrames.jl is that the functions in the `src => fun => dest` syntax act on vanilla vector inputs. We don't want to be tidyverse and have everyone use a DSL. 
4. I agree that the current proposal using `Vector`s only is not the best solution. However (1) We want to build an initial API that seems to work before trying to solve every other problem at once and (2) The current implementation of `View`s limits us to vectors.

CameronBieganek:
> 2\. We need an easy syntax so users can analyze data and be productive.

I suspect that `spreadmissings` is complex enough that it won't see much adoption from typical data analysts.

> Plus, I'm sure there are lots of instances where package authors justifiably want to work with a Vector or want to avoid thinking about missings. We provide a bridge between the data analyst and those packages.

You're basically splitting the ecosystem in half here. `spreadmissings` won't work for anyone who wants to input iterators into their function. For example, this won't work:

```julia
summeans(x, y) = mean(x) + mean(y)
x = Iterators.take([missing, 2, 3, 4], 3)
y = Iterators.take([5, missing, 6, 7], 3)
spreadmissings(summeans)(x, y)
```

As to the question of whether the post-processed inputs should be an iterator or a vector... It's not uncommon for users to be required to collect an iterator in certain places. For example, if you want to `sort` an iterator you have to collect it first. There might be a natural way to design the API so that users can `collect` if they need to. Or if nothing else, perhaps an `iterators` keyword argument to `spreadmissings` that defaults to `true`.

How about the following syntax, which works for inputs of arbitrary types?

```julia
foo(x, y, z) = z * (mean(x) + mean(y))

skipm(foo)(x, y, Ign(z))
```

or

```julia
struct NoMissings{T}
    x::T
end

const ¬ = NoMissings

skipm(foo)(x, y, ¬z)
```

As far as functions like `zscore` go, I would probably just write my own missing-skipping version. Or perhaps it would make sense for functions like `zscore` to have a keyword argument to control the missing behavior, like this:

```julia
function zscore(x; skipmissings=true)
    f = skipmissings ? skipmissing : identity
    (x .- mean(f(x))) ./ std(f(x))
end
```

One nice thing about this particular approach is that `zscore` can take an iterator as input:

```julia
julia> itr = Iterators.take([1, 2, 3, missing, 5], 4)
Base.Iterators.Take{Vector{Union{Missing, Int64}}}(Union{Missing, Int64}[1, 2, 3, missing, 5], 4)

julia> zscore(itr)
4-element Vector{Union{Missing, Float64}}:
 -1.0
  0.0
  1.0
   missing
```

But perhaps there's a more elegant way to do this. I will continue to think about it.

------

To clarify, my vision for the `skipm` function I used above is that it is basically this,

```julia
skipm(f) = (args...; kwargs...) -> f(skipmissings(args...)...; kwargs...)
```

except that it has extra handling to deal with arguments that are wrapped in `NoMissings`. So, `skipm` would handle `mean(x)`, `mean(f, x)`, `sum`, `var`, `cor`, `quantile`, `reduce`, etc. The only thing it doesn't handle is functions like `center` or `zscore`.

pdeffebach:
Sure, a keyword argument to control iteration vs `SubArray`  seems feasible. Maybe we can default to `skipmissings` when the inputs aren't all vectors. 

But I don't think you've responded to the scenario with DataFrames? A key benefit of the current approach is that you don't lose length information, the way you would with `skipmissings`. 

Also, keyword arguments are tough because

1. You need to convince package authors to add keyword arguments themselves
2. They don't do much more than dispatch or function-wrappers can.

CameronBieganek:
(Quoting myself.)

> There might be a natural way to design the API so that users can `collect` if they need to.

Of course, `skipmissings` is already fairly natural in this regard. If you need to collect the inputs to `bar`, you can do this:

```julia
bar(collect.(skipmissings(x, y))...)
```

> But I don't think you've responded to the scenario with DataFrames? A key benefit of the current approach is that you don't lose length information, the way you would with `skipmissings`.

Well, to be honest, I'm more interested in reduction functions. I feel like the need for a uniform missing-skipping syntax for reduction functions is more pressing than the need to handle functions like `center` and `zscore`.

But I did address the "DataFrames scenario" with non-reduction functions. I proposed either writing them by hand, like this,

```julia
transform(df, :x => (x -> x .- mean(skipmissing(x))) => :x_c)
```

or adding keyword arguments to `zscore`, etc. I also said "But perhaps there's a more elegant way to do this. I will continue to think about it."

I think trying to combine the behavior for reduction functions and non-reduction functions into one function produces a complicated, hard to understand function.

jariji:
Making `cor` accept `SkipMissings` would make 

```jl
cor(collect.(skipmissings(x, y))...)
```

look a little nicer without the `collect`.

pdeffebach:
> Making `cor` accept `SkipMissings` would make
> 
> ```julia
> cor(collect.(skipmissings(x, y))...)
> ```

Base devs rejected that proposal. see [here](https://github.com/JuliaStats/Statistics.jl/pull/30). Looking back, I agree with the logic a bit more. Vectors are a natural way to enforce indices matching, whereas iterators will never be as clear.

CameronBieganek:
Yeah, it's debatable. But it seems like they might be open to a new method like `cor(itr)`, where `itr` is an iterator of iterators.

CameronBieganek:
Copying over from Discourse... It's a little goofy, but we could signal to `skipm` that an argument needs to be collected. Something like this:

Setup:

```julia
struct NeedsArray{T}
    x::T
end

const ⋆ = NeedsArray
```

```julia
skipm(cor)(⋆x, ⋆y)
```

CameronBieganek:
Wow, the `cor(x::AbstractVector) = 1` method blows up any chance of having a `cor(itr)` method, where `itr` iterates observations. :sob:

In view of that unfortunate situation, perhaps the Base devs could be convinced to allow `cor(itr1, itr2)`.

nalimilan:
@CameronBieganek I think you should develop your use cases and/or make a detailed proposal somewhere else. You have hijacked a pull request about a well-defined function which was in the final review phase. This is not a design issue intended at experimenting with various ideas. The result is a confusing discussion which hasn't allowed me to grasp what are your main requirements.

CameronBieganek:
Apologies for hijacking a pull request. I have concerns about the design and usability of `spreadmissings`, which led me to suggest alternatives.

src/Missings.jl
```diff
@@ -208,6 +208,290 @@ missing
...
+    end
+
+    return out
+end
+
+function check_indices_match(vecs...)
+    Base.require_one_based_indexing(vecs...)
+    findex = eachindex(first(vecs))
+    # If vectors don't have the same indices, throw a
+    # nice error for the user.
```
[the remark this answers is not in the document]
pdeffebach: Thanks for this. I didn't know about it. I'll keep the `check_indices_match` function for readability.

src/Missings.jl
```diff
@@ -208,6 +208,290 @@ missing
...
+        if !(spread in (:default, :nonmissing, :none))
+            throw(ArgumentError("spread must be either :default, :nonmissing, or :none"))
+        end
+        new{Core.Typeof(f)}(f, spread)
+    end
+end
+
+function non_spreadabble_check(t::Union{AbstractDict, NamedTuple, Tuple})
+    T = typeof(t)
+    s = "Spreadmissings on $T is reserved. Please wrap in Ref to be " *
```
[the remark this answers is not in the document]
pdeffebach: No, it will not work... I will make it error for now.

src/Missings.jl
```diff
@@ -208,6 +208,290 @@ missing
...
+"""
+    spreadmissings(f; spread = :default)
+
+Given a function `f`, function `f` but performs a transformation
+on arguments to remove missing values before executing.
+
+### Initial example
+
+```julia-repl
+julia> using Statistics;
```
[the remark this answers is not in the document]
pdeffebach: Added.

src/Missings.jl
```diff
@@ -208,6 +208,290 @@ missing
...
+    elements of the inputs.
+* `:none`: `output` is returned directly, whether a `Vector` or not.
+
+A summary of the behavior is given in the table below:
+
+| spread \\ output type  | Vector           | Non-Vector       |
+|:---------------------- |:---------------- |:---------------- |
+| :default               | spread and match | return           |
+| :nonmissing            | spread and match | spread and match |
+| :none                  | return           | return           |
```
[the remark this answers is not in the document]
pdeffebach: The "match" bit is relevant when discussing the "spreading" behavior. Compare the way `@rtransform df :m = mean(:x)` to `@transform df :m = @spreadmissings mean(:x)` or something. In the former, you get the mean everywhere, in the latter you get the mean for non-missing indices only. It "matches" the output with the nonmissing indices. 

I added a note about this in the docstring. We should consider adding a keyword argument for this behavior eventually.

src/Missings.jl
```diff
@@ -208,6 +208,290 @@ missing
...
+to not include any missing values, return a `SubArray` referencing
+the `nonmissinginds`. The element type of the returned output
+does not include `missing`.
+"""
+function nomissing_subarray(a::AbstractVector, nonmissinginds::AbstractVector)
+    T = nonmissingtype(eltype(a)) # Element type
+    N = 1 # Dimension of view
+    P = typeof(a) # Type of parent array
+    I = Tuple{typeof(nonmissinginds)} # Type of the non-missing indices
+    L = Base.IndexStyle(a) == IndexLinear # If the type supports fast linear indexing (assumed true)
```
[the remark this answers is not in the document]
pdeffebach: added.

src/Missings.jl
```diff
@@ -208,6 +208,290 @@ missing
...
+        a = args[i]
+        if a isa AbstractVector
+            nomissing_subarray(a, nonmissinginds)
+        else
+            a
+        end
+    end
+end
+
+function maybespread_missing(f, newargs, new_kwargs, vecs, nonmissinginds, nonmissingmask)
```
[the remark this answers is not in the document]
pdeffebach: Added.

pdeffebach:
@nalimilan I have fixed all the issues you have highlighted. Thank you for your feedback!

Three things to keep in mind

1. `spreadmissings` currently hurts type stability. 
    
  ```
  julia> using Statistics
  
  julia> using Statistics, Missings
  
  julia> function foo(x)
             spreadmissings(mean)(x)
         end;
  
  julia> @code_warntype foo([1, missing])
  MethodInstance for foo(::Vector{Union{Missing, Int64}})
    from foo(x) @ Main REPL[27]:1
  Arguments
    #self#::Core.Const(foo)
    x::Vector{Union{Missing, Int64}}
  Body::Any
  1 ─ %1 = Main.spreadmissings(Main.mean)::Core.Const(Missings.SpreadMissings{typeof(mean)}(Statistics.mean, :default))
  │   %2 = (%1)(x)::Any
  └──      return %2
```

  I've determined this is *not* due to the fact that `spread` is a `Symbol` (as opposed to specialized structs) and *not* due to the fact that sometimes we return a vector without `missing`, in the case of no missing values in the inputs. 

  How much should I worry about this? Is there anything we can do? Maybe I could re-write the function so it dispatches on `SpreadMIssings{F, spread}`? 

2. I have added an error if `missing` is a positional or keyword argument. We can relax this in the future and propagate `missing` but I'm a little nervous about adding that behavior right now.

**Comment by nalimilan** [comment]
Have you tried adding `@inlined` to `spreadmissings(f; spread::Symbol = :default)` (and maybe `@noinline` to others)? At least that approach works for `levels` in DataAPI. You could also try using `if spread === :default; SpreadMissings(f, SpreadDefault())` and so on inside that function.

`spread=:all` could be useful indeed. I just hope things don't get too complex to grasp.

**Comment by jariji** [comment]
> I just hope things don't get too complex to grasp.

I'm well past that point. It would be easier for me to follow if each of these methods had its own function name rather than dispatching on keyword arguments.

**Comment by pdeffebach** [comment]
@nalimilan this is ready for a review!

The inference failures were due to a lack of function barriers. Adding `spread` to the type parameter of `SpreadMissings` and using dispatch to control behavior rather than `if...else` works well. Type hints in the smaller functions also appear helpful. 

I've filed an issue upstream for a lack of type inference with `mean`. See [here](https://github.com/JuliaStats/Statistics.jl/issues/160). The `SubArray` hack does not go as far as I would like, and still produces a lot of `Union{Missing, Float64}` inference. But that's okay since it's a small union. 

I also added `spread = :all`

* For scalar outputs, spreads the output along the full length of the input vector
* For vector outputs, throws an error. This is because the inner function knows nothing about the full length of the input vectors to the outer function. Any scenario where the output vector matches the length of the input vectors is coincidence and the indices wouldn't correspond. 

It's worth emphasizing again that if none of the inputs are vectors, then behavior is entirely unaffected. So no error is thrown if the inputs are not vectors and the output is a vector. 

This behavior is indeed complicated, but it's all in the docstring.

**Comment by pdeffebach** [comment]
@nalimilan Bumping this. Remember that we ultimately want to try this out in DataFramesMeta.jl first. I will port the PR over there eventually but am editing it here.

File: src/Missings.jl [hunk 23]
```diff
@@ -208,6 +208,447 @@ missing
...
+    spread::SpreadDefault,
+    vecs::Tuple)::typeof(res) where {T}
+
+    spread_nomissing(res, vecs)
+end
+
+function maybespread_nomissing(
+    res::T,
+    spread::SpreadNonMissing,
+    vecs::Tuple)::Vector{T} where{T}
```
nalimilan, line 383: Are all these return type annotations needed? In many cases they seem redundant.

src/Missings.jl
```diff
@@ -208,6 +208,447 @@ missing
...
+    vecs::Tuple,
+    nonmissinginds::AbstractVector{<:Integer},
+    nonmissingmask::AbstractVector{<:Bool})
+
+    throw(ArgumentError("spreadmissings with :all on vector output is reserved"))
+end
+
+function spread_nomissing(
+    res::AbstractVector{T},
+    vecs::Tuple)::typeof(res) where {T}
```
nalimilan, line 352: For clarity, I would make this `maybespread_nomissing(res::AbstractVector{T}, spread::SpreadNonMissing, vecs::Tuple)`. Same for `maybespread_missing`.

src/Missings.jl
```diff
@@ -208,6 +208,447 @@ missing
...
+    res
+end
+
+function maybespread_nomissing(
+    res::T,
+    spread::SpreadAll,
+    vecs::Tuple)::AbstractVector{T} where {T}
+
+    out = Vector{typeof(res)}(undef, length(first(vecs)))
+    out .= Ref(res)
```
nalimilan, line 412: How about calling `maybspread_nomissing(res, SpreadNonMissing(), vecs)`?

src/Missings.jl
```diff
@@ -208,6 +208,447 @@ missing
...
+struct SpreadNone <: AbstractSpread end
+struct SpreadAll <: AbstractSpread end
+
+struct SpreadMissings{F, S <: AbstractSpread} <: Function
+    f::F
+    spread::S
+    function SpreadMissings(f, spread::AbstractSpread)
+        if !(spread isa AbstractSpread)
+            throw(ArgumentError("spread must be either SpreadDefault(), SpreadNonMissing(), or SpreadNone()"))
+        end
```
nalimilan, line 223: This cannot happen.
```suggestion
```

src/Missings.jl
```diff
@@ -208,6 +208,447 @@ missing
...
+      missing
+    ```
+
+    1. `fillmean_skip` fills all entries of the original vector `x` with the mean,
+       excluding `missing`s. By contrast, `spreadmissings(fillmean)` only fills non-missing
+       elements of the original `x`.
+    2. `fillmean_skip` returns a vector which does not allow for `missing`, while
+       `spreadmissings(fillmean)` does.
+
+    Use the keyword `spread = :all` to emulate the `skipmissing` behavior.
```
nalimilan, line 636: Show an example?

test/.ipynb_checkpoints/runtests-checkpoint.jl
```diff
@@ -0,0 +1,227 @@
+using Test, SparseArrays, Missings
```
nalimilan, line 1: What's this file?

test/spreadmissings.jl
```diff
@@ -0,0 +1,212 @@
...
+const ≈ = isequal
+
+function small_vec(args...; kwargs...)
+    [1, 2]
+end
+
+function right_vec(args...; kwargs...)
+    kwargs_vals = values(values(kwargs))
+    xs = tuple(args..., kwargs_vals...)
+    vecs = Base.filter(x -> x isa AbstractVector, xs)
```
nalimilan, line 14:
```suggestion
    vecs = filter(x -> x isa AbstractVector, xs)
```

test/spreadmissings.jl
```diff
@@ -31,119 +31,176 @@ y = [100, 200, 300, 400]
...
+    x = [1, 2, 3]
+    t = spreadmissings(categorical)(x)
+    @test t == categorical([1, 2, 3])
+    @test typeof(t) == typeof(categorical([1, 2, 3]))
+
+    x = [1, 2, 3, missing]
+    t = spreadmissings(categorical)(x)
+    @test t ≈ categorical([1, 2, 3, missing])
+    @test typeof(t) == typeof(categorical([1, 2, 3, missing]))
+end
```
nalimilan, line 204: Can you add tests that inference works?

src/Missings.jl
```diff
@@ -208,6 +208,448 @@ missing
...
+the `nonmissinginds`. The element type of the returned output
+does not include `missing`.
+"""
+function nomissing_subarray(a::AbstractVector, nonmissinginds::AbstractVector)
+    T = nonmissingtype(eltype(a)) # Element type
+    N = 1 # Dimension of view
+    P = typeof(a) # Type of parent array
+    I = Tuple{typeof(nonmissinginds)} # Type of the non-missing indices
+    L = Base.IndexStyle(a) === IndexLinear # If the type supports fast linear indexing
+    SubArray{T, N, P, I, L}(a, (nonmissinginds,), 0, 1)
```
CameronBieganek, line 242: The docstring for `SubArray` makes no mention of contructors, and in fact says "Construct `SubArray`s using the `view` function." So this direct usage of a `SubArray` constructor is usage of undocumented Base internals that could change in any minor or patch release.

pdeffebach: Yeah, that's probably true. 

I think it will be worthwhile to make our own array type that does this better... hopefully it's not too hard.

PR #126, opened 2021-01-24
Supporting indexing EachReplaceMissing and EachFailMissing

nalimilan:
Similar to `SkipMissing`.

(I need this to support missing values at https://github.com/JuliaData/DataFrames.jl/pull/2610.)


nalimilan:
Failure on nightly is https://github.com/JuliaData/DataAPI.jl/pull/32.

src/Missings.jl
```diff
@@ -87,6 +92,8 @@ Base.IteratorEltype(::Type{<:EachReplaceMissing{T}}) where {T} =
 Base.length(itr::EachReplaceMissing) = length(itr.x)
 Base.size(itr::EachReplaceMissing) = size(itr.x)
 Base.eltype(itr::EachReplaceMissing) = nonmissingtype(eltype(itr.x))
+Base.eachindex(itr::EachReplaceMissing) = eachindex(itr.x)
+Base.keys(itr::EachReplaceMissing) = keys(itr.x)
```
bkamins, line 96: These two will fail if `itr.x` does not define them, but maybe it is OK.

src/Missings.jl
```diff
@@ -102,11 +109,19 @@ end
     return (v isa Missing ? itr.replacement : v, s)
 end
 
+Base.@propagate_inbounds function Base.getindex(itr::EachReplaceMissing, I...)
```
bkamins, line 112: same here.

bkamins (approved):

bkamins:
Just please check the failure of CI on nightly.

src/Missings.jl
nalimilan: Yes that's intended. We do the same for `SkipMissing` in Base.

bkamins: so maybe we should support `length`, `size` etc. (all the array interface) in the same way? We cannot do it for `SkipMissing` but here we can.

nalimilan: We already do. :-)

nalimilan: Ah, we need `axes`. Will add it.

nalimilan:
CI failure on nightly is fixed by https://github.com/JuliaData/Missings.jl/pull/127.

nalimilan:
https://github.com/JuliaRegistries/General/pull/28600

Issue #142, opened 2022-02-17
Add `isless` variant where `missing` is smallest


bkamins:
The definitions to make `missing` smallest would be
```
isless2(::Missing, ::Missing) = false
isless2(::Missing, ::Any) = true
isless2(::Any, ::Missing) = false
isless2(x, y) = isless(x, y)
```

@nalimilan - the question is what should be the name of this function (as usual this is the hardest part).

Also a question is if Missings.jl should export it (I think it is OK to export it if we find a good name)

nalimilan:
`islessmissingsmallest` or `islessmissingfirst` would be explicit though a bit long...

bkamins:
This is my fear. Let us ask on Slack.

sprmnt21:
let's start by removing `isless`, since it is the default value 
`missingassmallest`
`missingissmallest`

sprmnt21:
it sounds like a play on words (or is it? I like it).
`missingless `could be read as _no missing_

bkamins:
The approach we are discussing is more general and would allow you to sort any collection, e.g. a vector in exactly the same way.

nalimilan:
Any objections to `missingisless`?

BTW, since the main use case is to do `sort(x, lt=missingisless, rev=true)`, we could directly provide a function which reverses the order. But I don't have a good name for that (`missingisnotgreater`?)

bkamins:
`missingisnotgreater` feels unclear.

`missingisless` seems OK.

PR #144, opened 2023-07-19
Implementing order functions where missings is the smallest value

alonsoC1s:
As discussed in issue #142, this PR adds a variant of the standard `isless` function where the smallest possible value is `missing`. Furthermore, this PR adds a partial order function `missingsmallest` that takes a partial order function `f` and modifies the order such that `missing` is always less than the other argument.

This PR aims to provide support to resolve issue https://github.com/JuliaData/DataFrames.jl/issues/2267 at DataFrames.jl

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,8 @@ This package provides additional functionality for working with `missing` values
 - `Missings.replace` to wrap a collection in a (possibly indexable) iterator replacing `missing` with another value
 - `Missings.fail` to wrap a collection in a (possibly indexable) iterator throwing an error if `missing` is encountered
 - `skipmissings` to loop through a collection of iterators excluding indices where any iterators are `missing`
-- `missingsmallest(f)` to create a partial order function that behaves as `f` and making `missing` the smallest value
-- `missingsless` the standard `isless` function modified so that `missing` is less than any other element
+- `missingsmallest(f)` to create a partial order function that treats `missing` as the smallest value and otherwise behaves like `f`
+- `missingsless` the standard `isless` function modified to treat `missing` as the smallest value rather than the largest one
 
 ## Contributing and Questions
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ This package provides additional functionality for working with `missing` values
 - `Missings.replace` to wrap a collection in a (possibly indexable) iterator replacing `missing` with another value
 - `Missings.fail` to wrap a collection in a (possibly indexable) iterator throwing an error if `missing` is encountered
 - `skipmissings` to loop through a collection of iterators excluding indices where any iterators are `missing`
+- `missingsmallest(f)` to create a partial order function that behaves as `f` and making `missing` the smallest value
+- `missingsless` the standard `isless` function modified so that `missing` is less than any other element
 
 ## Contributing and Questions
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ This package provides additional functionality for working with `missing` values
 - `Missings.fail` to wrap a collection in a (possibly indexable) iterator throwing an error if `missing` is encountered
 - `skipmissings` to loop through a collection of iterators excluding indices where any iterators are `missing`
 - `missingsmallest(f)` to create a partial order function that treats `missing` as the smallest value and otherwise behaves like `f`
-- `missingsmallest` the standard `isless` function modified to treat `missing` as the smallest value rather than the largest one
+- `missingsmallest`: the standard `isless` function modified to treat `missing` as the smallest value rather than the largest one
 
 ## Contributing and Questions
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ This package provides additional functionality for working with `missing` values
 - `Missings.fail` to wrap a collection in a (possibly indexable) iterator throwing an error if `missing` is encountered
 - `skipmissings` to loop through a collection of iterators excluding indices where any iterators are `missing`
 - `missingsmallest(f)` to create a partial order function that treats `missing` as the smallest value and otherwise behaves like `f`
-- `missingsless` the standard `isless` function modified to treat `missing` as the smallest value rather than the largest one
+- `missingsmallest` the standard `isless` function modified to treat `missing` as the smallest value rather than the largest one
 
 ## Contributing and Questions
 
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -534,8 +534,7 @@ y)` function. This is equivalent to `missingsmallest(isless)(x, y)`.
 
 # Examples
 ```
-julia> lengthmissing = passmissing(length)
-julia> isshorter = missingsmallest((s1, s2) -> isless(lengthmissing(s1), lengthmissing(s2)))
+julia> isshorter = missingsmallest((s1, s2) -> isless(length(s1), length(s2)))
 julia> isshorter("short", "longstring")
 true
 
@@ -554,8 +553,8 @@ missingsmallest(f) = MissingSmallest(f)
 The standard partial order `isless` modified so that `missing` is always the
 smallest possible value:
 - If neither argument is `missing`, the function behaves exactly as `isless`.
-- If `x` is `missing` the result will be `true` regardless of the value of `y`.
 - If `y` is `missing` the result will be `false` regardless of the value of `x`.
+- If `x` is `missing` the result will be `true` unless `y` is `missing`.
 
 See also the 1-argument method which takes a partial ordering function (like
 `isless`) and modifies it to treat `missing` as explained above. These functions
@@ -564,8 +563,7 @@ first. This is useful in particular so that when sorting in reverse order
 missing values appear at the end.
 
 # Examples
-```
-julia> v = [missing, 10, missing, 1, 2]
+```jldoctest
 julia> sort(v, lt=missingsmallest)
 
 5-element Vector{Union{Missing, Int64}}:
@@ -591,8 +589,7 @@ julia> missingsmallest(-Inf, missing)
 false
 
 julia> missingsmallest(missing, missing)
-true
-```
+false
 """
 missingsmallest(x, y) = missingsmallest(isless)(x, y)
 
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -528,17 +528,21 @@ words, return a modified version of the partial order function `f` such that
 `missing` is the smallest possible value, and all other non-`missing` values are
 compared according to `f`.
 
-The traditional `isless` function modified to treat `missing` as the smallest
-value can be accessed by using the 2-argument `missingsmallest(x, y)` function,
-see [`missingsmallest`](@ref).
+The behavior of the standard `isless` function modified to treat `missing` as
+the smallest value can be obtained by calling the 2-argument `missingsmallest(x,
+y)` function. This is equivalent to `missingsmallest(isless)(x, y)`.
 
 # Examples
 ```
-julia> ismuchless = missingsmallest((x, y) -> isless(x, 10*y))
-julia> ismuchless(1, 100)
+julia> lengthmissing = passmissing(length)
+julia> isshorter = missingsmallest((s1, s2) -> isless(lengthmissing(s1), lengthmissing(s2)))
+julia> isshorter("short", "longstring")
 true
 
-julia> ismuchless(-Inf, missing)
+julia> isshorter("longstring", "short")
+false
+
+julia> isshorter("", missing) # Is shorter than length 0?
 false
 ```
 """
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -528,12 +528,17 @@ words, returns a modified version of the partial order function `f` such that
 `missing` is the smallest possible value, and all other non-`missing` values are
 compared according to `f`.
 
+The traditional `isless` function modified to treat `missing` as the smallest
+value can be accessed by using the 2-argument `missingsmallest(x, y)` function,
+see [`missingsmallest`](@ref).
+
 # Examples
 ```
-julia> missingsmallest((x, y) -> isless(x, 10*y))(missing, Inf)
+julia> ismuchless = missingsmallest((x, y) -> isless(x, 10*y))
+julia> ismuchless(1, 100)
 true
 
-julia> missingsless(-Inf, missing)
+julia> ismuchless(-Inf, missing)
 false
 ```
 """
@@ -548,10 +553,10 @@ smallest possible value. The expected behaviour is the following:
 - If `x` is `missing` the result will be `true` regardless of the value of `y`.
 - If `y` is `missing` the result will be `false` regardless of the value of `x`.
 
-See also [`missingsmallest`](@ref), which is equivalent to using
-`missingsmallest(isless)`. These functions can be used together with sorting
-functions such that the first elements of the newly sorted Array are placed
-first.
+See also [`missingsmallest`](@ref), the 1-argument function which takes a
+partial ordering function (like `isless`) and modifies it to treat `missing` as
+explained above. These functions can be used together with sorting functions
+such that the first elements of the newly sorted Array are placed first.
 
 # Examples
 ```
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -2,7 +2,7 @@ module Missings
 
 export allowmissing, disallowmissing, ismissing, missing, missings,
        Missing, MissingException, levels, coalesce, passmissing, nonmissingtype,
-       skipmissings, emptymissing
+       skipmissings, emptymissing, missingsmallest, missingsless
 
 using Base: ismissing, missing, Missing, MissingException
 
@@ -513,5 +513,33 @@ julia> emptymissing(first)([1], 2)
 ```
 """
 emptymissing(f) = (x, args...; kwargs...) -> isempty(x) ? missing : f(x, args...; kwargs...)
+
+# Variant of `isless` where `missing` is the smallest value
+"""
+    missingsmallest(f::Function)
+
+Creates a function of two arguments `x` and `y` that tests whether `x` is less
+than `y` such that `missing` is always less than the other argument. In other
+words, modifies the partial order function `f` such that `missing` is the
+smallest possible value, and all other non-`missing` values are compared
+according to `f`.
+
+See also: [`missingsless`](@ref), the standard order where `missing` is the
+smallest possible value.
+
+# Examples
+```
+julia> missingsmallest(isless)(missing, Inf)
+true
+
+julia> missingsless(-Inf, missing)
+false
+```
+"""
+missingsmallest(f) = (x, y) -> ismissing(y) ? false : ismissing(x) ? true : f(x, y)
+
+" The standard partial order `isless` modified so that `missing` is always the
+smallest possible value."
+const missingsless = missingsmallest(isless)
 
 end # module
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -539,10 +539,6 @@ false
 """
 missingsmallest(f) = MissingSmallest(f)
 
-
-" The standard partial order `isless` modified so that `missing` is always the
-smallest possible value."
-
 """
     missingsmallest(x, y)
 
@@ -552,11 +548,22 @@ smallest possible value. The expected behaviour is the following:
 - If `x` is `missing` the result will be `true` regardless of the value of `y`.
 - If `y` is `missing` the result will be `false` regardless of the value of `x`.
 
-See also [`missingsmallest`](@ref), which modifies a partial order function and
-yields a function that behaves as the expected behaviour outlined above.
+See also [`missingsmallest`](@ref), which is equivalent to using
+`missingsmallest(isless)`. These functions can be used together with sorting
+functions such that the first elements of the newly sorted Array are placed
+first.
 
 # Examples
 ```
+julia> v = [missing, 10, missing, 1, 2]
+julia> sort(v, lt=missingsmallest)
+5-element Vector{Union{Missing, Int64}}:
+   missing
+   missing
+  1
+  2
+ 10
+
 julia> missingsmallest(missing, Inf)
 true
 
@@ -570,21 +577,5 @@ true
 missingsmallest(x, y) = missingsmallest(isless)(x, y)
 
 (ms::MissingSmallest)(x, y) = ismissing(y) ? false : ismissing(x) ? true : ms.lt(x, y)
-
-# Some optimizations
-const _MissingSmallest = Union{MissingSmallest, typeof(missingsmallest)}
-
-@static if isdefined(Base.Sort, :MissingOptimization) && isdefined(Base.Sort, :_sort!)
-    function Base.Sort._sort!(v::AbstractVector, a::Base.Sort.MissingOptimization,
-        o::Base.Order.Lt{<:_MissingSmallest}, kw)
-        # put missing at beginning
-        Base.Sort._sort!(v, a.next, withoutmissingordering(o), kw...;hi=new_hi)
-    end
-    function Base.Sort._sort!(v::AbstractVector, a::Base.Sort.MissingOptimization,
-        o::Base.Order.ReverseOrdering{Base.Order.Lt{<:_MissingSmallest}}, kw)
-        # put missing at end
-        Base.Sort._sort!(v, a.next, ReverseOrdering(withoutmissingordering(o.fwd)), kw...; lo=new_lo)
-    end
-end
 
 end # module
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -541,7 +541,7 @@ true
 julia> isshorter("longstring", "short")
 false
 
-julia> isshorter("", missing) # Is shorter than length 0?
+julia> isshorter("", missing)
 false
 ```
 """
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -534,7 +534,8 @@ y)` function. This is equivalent to `missingsmallest(isless)(x, y)`.
 
 # Examples
 ```
-julia> isshorter = missingsmallest((s1, s2) -> isless(length(s1), length(s2)))
+julia> isshorter = missingsmallest((s1, s2) -> isless(length(s1), length(s2)));
+
 julia> isshorter("short", "longstring")
 true
 
@@ -574,7 +575,6 @@ julia> sort(v, lt=missingsmallest)
  10
 
 julia> sort(v, lt=missingsmallest, rev=true)
-
 5-element Vector{Union{Missing, Int64}}:
  10
   2
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -518,26 +518,51 @@ emptymissing(f) = (x, args...; kwargs...) -> isempty(x) ? missing : f(x, args...
     missingsmallest(f)
 
 Creates a function of two arguments `x` and `y` that tests whether `x` is less
-than `y` such that `missing` is always less than the other argument. In other
-words, modifies the partial order function `f` such that `missing` is the
-smallest possible value, and all other non-`missing` values are compared
-according to `f`.
+than `y` such that `missing` is always less than the other argument. In other 
+words, returns a modified version of the partial order function `f` such that
+`missing` is the smallest possible value, and all other non-`missing` values are
+compared according to `f`.
 
 See also: [`missingsless`](@ref), equivalent to `missingsmallest(isless)`
 
 # Examples
 ```
-julia> missingsmallest(isless)(missing, Inf)
-true
+julia> missingsmallest(Base.isgreater)(missing, Inf)
+false
 
 julia> missingsless(-Inf, missing)
 false
 ```
 """
-missingsmallest(f) = (x, y) -> ismissing(y) ? false : ismissing(x) ? true : f(x, y)
+@inline missingsmallest(f) = (x, y) -> ismissing(y) ? false : ismissing(x) ? true : f(x, y)
 
 " The standard partial order `isless` modified so that `missing` is always the
 smallest possible value."
-const missingsless = missingsmallest(isless)
+
+"""
+    missingsless(x, y)
+
+The standard partial order `isless` modified so that `missing` is always the
+smallest possible value. The expected behaviour is the following:
+- If neither argument is `missing`, the function behaves exactly as `isless`.
+- If `x` is `missing` the result will be `true` regardless of the value of `y`.
+- If `y` is `missing` the result will be `false` regardless of the value of `x`.
+
+See also [`missingsmallest`](@ref), which modifies a partial order function and
+yields a function that behaves as the expected behaviour outlined above.
+
+# Examples
+```
+julia> missingsless(missing, Inf)
+true
+
+julia> missingsless(-Inf, missing)
+false
+
+julia> missingsless(missing, missing)
+true
+```
+"""
+missingsless(x, y) = missingsmallest(isless)(x, y)
 
 end # module
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -566,7 +566,6 @@ missing values appear at the end.
 # Examples
 ```jldoctest
 julia> sort(v, lt=missingsmallest)
-
 5-element Vector{Union{Missing, Int64}}:
    missing
    missing
--- a/src/Missings.jl
+++ b/src/Missings.jl
@@ -514,9 +514,8 @@ julia> emptymissing(first)([1], 2)
 """
 emptymissing(f) = (x, args...; kwargs...) -> isempty(x) ? missing : f(x, args...; kwargs...)
 
-# Variant of `isless` where `missing` is the smallest value
 """
-    missingsmallest(f::Function)
+    missingsmallest(f)
 
 Creates a function of two arguments `x` and `y` that tests whether `x` is less
 than `y` such that `missing` is always less than the other argument. In other
@@ -524,8 +523,7 @@ words, modifies the partial order function `f` such that `missing` is the
 smallest possible value, and all other non-`missing` values are compared
 according to `f`.
 
-See also: [`missingsless`](@ref), the standard order where `missing` is the
-smallest possible value.
+See also: [`missingsless`](@ref), equivalent to `missingsmallest(isless)`
 
 # Examples
 ```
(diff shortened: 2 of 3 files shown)


src/Missings.jl
```diff
@@ -514,4 +514,32 @@ julia> emptymissing(first)([1], 2)
 """
 emptymissing(f) = (x, args...; kwargs...) -> isempty(x) ? missing : f(x, args...; kwargs...)
 
+# Variant of `isless` where `missing` is the smallest value
+"""
+    missingsmallest(f::Function)
+
+Creates a function of two arguments `x` and `y` that tests whether `x` is less
```
bkamins, line 521: Returns?

src/Missings.jl
```diff
@@ -514,4 +514,32 @@ julia> emptymissing(first)([1], 2)
...
+than `y` such that `missing` is always less than the other argument. In other
```
bkamins, line 522: and otherwise `f` is used to perform the comparison.

src/Missings.jl
```diff
@@ -514,4 +514,32 @@ julia> emptymissing(first)([1], 2)
...
+# Examples
+```
+julia> missingsmallest(isless)(missing, Inf)
+true
+
+julia> missingsless(-Inf, missing)
+false
+```
+"""
+missingsmallest(f) = (x, y) -> ismissing(y) ? false : ismissing(x) ? true : f(x, y)
```
bkamins, line 539: `@inline`?

src/Missings.jl
```diff
@@ -514,4 +514,32 @@ julia> emptymissing(first)([1], 2)
...
+
+" The standard partial order `isless` modified so that `missing` is always the
+smallest possible value."
+const missingsless = missingsmallest(isless)
```
bkamins, line 543: maybe it would be better to define:
```
missingsless(x, y) = missingsmallest(isless)(x, y)
```
Then the `typeof(missingsless)` will be more informative.

src/Missings.jl
```diff
@@ -514,4 +514,32 @@ julia> emptymissing(first)([1], 2)
...
+julia> missingsmallest(isless)(missing, Inf)
+true
+
+julia> missingsless(-Inf, missing)
+false
+```
+"""
+missingsmallest(f) = (x, y) -> ismissing(y) ? false : ismissing(x) ? true : f(x, y)
+
+" The standard partial order `isless` modified so that `missing` is always the
```
bkamins, line 541: missing call signature in docstring

bkamins:
I will ask users on slack for opinion about naming.

jariji:
`missingsleast` would be more consistent with the existing "less" convention. Perhaps that usage ("`missing` is the least of all the values") is uncommon/unfamiliar, however.

bkamins:
A good comment by Alexander Plavin on Slack:
```
julia> sort([1, missing, 3, 2], by=x->(!ismissing(x), x))
4-element Vector{Union{Missing, Int64}}:
  missing
 1
 2
 3
```
Maybe it is easy enough (and just should be documented)

The only downside is that in interactive use it triggers recompilation each time.

README.md
```diff
@@ -28,6 +28,8 @@ This package provides additional functionality for working with `missing` values
 - `Missings.replace` to wrap a collection in a (possibly indexable) iterator replacing `missing` with another value
 - `Missings.fail` to wrap a collection in a (possibly indexable) iterator throwing an error if `missing` is encountered
 - `skipmissings` to loop through a collection of iterators excluding indices where any iterators are `missing`
+- `missingsmallest(f)` to create a partial order function that behaves as `f` and making `missing` the smallest value
```
nalimilan, line 31:
```suggestion
- `missingsmallest(f)` to create a partial order function that treats `missing` as the smallest value and otherwise behaves like `f`
```

src/Missings.jl
```diff
@@ -514,4 +514,32 @@ julia> emptymissing(first)([1], 2)
 """
 emptymissing(f) = (x, args...; kwargs...) -> isempty(x) ? missing : f(x, args...; kwargs...)
 
+# Variant of `isless` where `missing` is the smallest value
+"""
+    missingsmallest(f::Function)
```
nalimilan, line 519: Technically, objects other than functions are allowed (e.g. types):
```suggestion
    missingsmallest(f)
```

src/Missings.jl
nalimilan: The convention in Julia is to use the imperative "Return". :-)

src/Missings.jl
```diff
@@ -514,4 +514,32 @@ julia> emptymissing(first)([1], 2)
...
+
+Creates a function of two arguments `x` and `y` that tests whether `x` is less
+than `y` such that `missing` is always less than the other argument. In other
+words, modifies the partial order function `f` such that `missing` is the
+smallest possible value, and all other non-`missing` values are compared
+according to `f`.
+
+See also: [`missingsless`](@ref), the standard order where `missing` is the
+smallest possible value.
```
nalimilan, line 528:
```suggestion
See also: [`missingsless`](@ref), equivalent to `missingsmallest(isless)`
```

src/Missings.jl
```diff
@@ -514,4 +514,32 @@ julia> emptymissing(first)([1], 2)
...
+
+# Examples
+```
+julia> missingsmallest(isless)(missing, Inf)
```
nalimilan, line 532: Show an example which isn't equivalent to `missingsless` instead? Otherwise it's not a good sign that we actually need this function at all.

test/runtests.jl
```diff
@@ -257,4 +257,11 @@ struct CubeRooter end
     @test emptymissing(fun)(3, 1, c=2) == (1, 2)
 end
 
+@testset "missingsmallest" begin
```
nalimilan, line 260: Need to test `missingsmallest` on a different comparison function.

test/runtests.jl
```diff
@@ -257,4 +257,11 @@ struct CubeRooter end
...
+    @test missingsless(missing, Inf) == true
+    @test missingsless(-Inf, missing) == false
```
nalimilan, line 262:
```suggestion
    @test missingsless(-Inf, missing) == false
    @test missingsless(missing, missing) == false
```

src/Missings.jl
nalimilan: And given that it will be the most commonly used function of the pair, it would be good to have a full docstring with examples and a reference to `missingsmallest`.

README.md
```diff
@@ -28,6 +28,8 @@ This package provides additional functionality for working with `missing` values
...
+- `missingsless` the standard `isless` function modified so that `missing` is less than any other element
```
nalimilan, line 32: To make the formulation more consistent with the previous one (which helps understanding the commonalities and differences):
```suggestion
- `missingsless` the standard `isless` function modified to treat `missing` as the smallest value rather than the largest one
```

src/Missings.jl
```diff
@@ -514,4 +514,32 @@ julia> emptymissing(first)([1], 2)
 """
 emptymissing(f) = (x, args...; kwargs...) -> isempty(x) ? missing : f(x, args...; kwargs...)
 
+# Variant of `isless` where `missing` is the smallest value
```
nalimilan, line 517: This doesn't describe `missingsmallest`. Not really useful anyway since we have a docstring.
```suggestion
```

nalimilan:
> Maybe it is easy enough (and just should be documented)

IMO it's too hard to discover and remember. At least `missingsless` seems useful to simplify the syntax.

Let's continue the discussion regarding names at https://github.com/JuliaData/DataFrames.jl/issues/2267 to avoid spreading it over several issues, and discuss the implementation here?

src/Missings.jl
alonsoC1s: I feel like this is already covered in the last sentence "... `missing` is the smallest possible value,
and all other non-`missing` values are compared according to `f`."

alonsoC1s:
While thinking of new test sets I came across something interesting. The current implementation only works for comparisons of the same "semantics" as `isless`. i.e comparisons where the first argument is tested as smallest, which is not the case for other comparison functions. For instance, `Base.isgreater`. Currently, we would expect `Inf` > `missing` to be `true` always. But since we rely on the order of the arguments, and `missing` is the second argument, this returns `false` by default.

I think this is either okay if we only expect users to compare using comparison functions that behave like `isless` (for instance, the "is much smaller" function) or not okay if we want this to work for any arbitrary function of two arguments that returns `bool`.

nalimilan:
> While thinking of new test sets I came across something interesting. The current implementation only works for comparisons of the same "semantics" as `isless`. i.e comparisons where the first argument is tested as smallest, which is not the case for other comparison functions. For instance, `Base.isgreater`. Currently, we would expect `Inf` > `missing` to be `true` always. But since we rely on the order of the arguments, and `missing` is the second argument, this returns `false` by default.

Could you show examples? I'm a bit confused about what you mean.

src/Missings.jl
nalimilan: It would be good for the first sentence to give a complete and accurate description of what happens. "is less" is not really accurate here since it's just the result of `f`. The interpretation of that result as meaning "is less" only appears when you use this function in combination with `sort`.

alonsoC1s:
Sorry, I think I could have explained it a bit better. What I mean is, the fact that we rely on the order of the arguments to place `missing` as the smallest value limits the usage of `missingsmallest` to ordering functions that are equivalent to `<` (or `isless`) and excludes functions equivalent to `>` (or `Base.isgreater`)

More concretely.
- `missing < x` should be `true` for any `x`.
- By the same argument `missing > x` should be `false` for any `x`

But since the current implementation _always_ returns `true` if the first argument is `missing`, `missing > x` would return `true`; there is no way to modify the order `>` such that `missing` is the smallest possible value.

As I said I am not sure if this could be a problem. In the concrete example of `Base.isgreater`, `missing` is already the smallest possible value as mentioned in their docstring. So this might not require a fix

bkamins:
Your comment is good. Maybe we do not need to add anything but instead recommend using the following:

```
julia> sort([1, missing, 3, 2])
4-element Vector{Union{Missing, Int64}}:
 1
 2
 3
  missing

julia> sort([1, missing, 3, 2], rev=true)
4-element Vector{Union{Missing, Int64}}:
  missing
 3
 2
 1

julia> sort([1, missing, 3, 2], lt=Base.isgreater)
4-element Vector{Union{Missing, Int64}}:
 3
 2
 1
  missing

julia> sort([1, missing, 3, 2], lt=Base.isgreater, rev=true)
4-element Vector{Union{Missing, Int64}}:
  missing
 1
 2
 3
```

and this is enough. Then the only question is why `Base.isgreater` is not exported by Base. Maybe we could export it in Missings.jl with the guarantee that if `Base.isgreater` is removed from Base we would provide an implementation.

`Base.isgreater` was added in https://github.com/JuliaLang/julia/pull/35316. @StefanKarpinski - you merged this change. So maybe you can comment on why `Base.isgreater` is not exported? Thank you!

alonsoC1s:
I think that would work. My only concern would be, what if the user wants to use something that behaves like `isgreater(x, y)` but not exactly that. For instance, the `ismuchgreater` function that returns `true` if `x > 100 * y`.  That case can't be solved by using `missingsmallest`, nor `Base.isgreater`.

bkamins:
@alonsoC1s - so we have a conclusion regarding the API in https://github.com/JuliaData/DataFrames.jl/issues/2267.
Do you have time to finalize this PR?

alonsoC1s:
@bkamins I implemented the version suggested by @LilithHafner without the the changes that would need to be upstreamed to Base

src/Missings.jl
```diff
@@ -514,4 +514,77 @@ julia> emptymissing(first)([1], 2)
...
+(ms::MissingSmallest)(x, y) = ismissing(y) ? false : ismissing(x) ? true : ms.lt(x, y)
+
+# Some optimizations
+const _MissingSmallest = Union{MissingSmallest, typeof(missingsmallest)}
+
+@static if isdefined(Base.Sort, :MissingOptimization) && isdefined(Base.Sort, :_sort!)
+    function Base.Sort._sort!(v::AbstractVector, a::Base.Sort.MissingOptimization,
+        o::Base.Order.Lt{<:_MissingSmallest}, kw)
+        # put missing at beginning
+        Base.Sort._sort!(v, a.next, withoutmissingordering(o), kw...;hi=new_hi)
```
LilithHafner, line 581: This is pseudocode, it needs to actually put the missing values at the beginning.

LilithHafner: Or just remove the optimizations.
And if there are optimizations, there should be tests that cover those branches.

src/Missings.jl
```diff
@@ -514,4 +514,77 @@ julia> emptymissing(first)([1], 2)
...
+words, returns a modified version of the partial order function `f` such that
+`missing` is the smallest possible value, and all other non-`missing` values are
+compared according to `f`.
+
+# Examples
+```
+julia> missingsmallest((x, y) -> isless(x, 10*y))(missing, Inf)
+true
+
+julia> missingsless(-Inf, missing)
```
bkamins, line 536: this is an example of other definition

src/Missings.jl
```diff
@@ -514,4 +514,77 @@ julia> emptymissing(first)([1], 2)
...
+    missingsmallest(x, y)
+
+The standard partial order `isless` modified so that `missing` is always the
+smallest possible value. The expected behaviour is the following:
+- If neither argument is `missing`, the function behaves exactly as `isless`.
+- If `x` is `missing` the result will be `true` regardless of the value of `y`.
+- If `y` is `missing` the result will be `false` regardless of the value of `x`.
+
+See also [`missingsmallest`](@ref), which modifies a partial order function and
+yields a function that behaves as the expected behaviour outlined above.
```
bkamins, line 556: maybe add that writing `missingsmallest` is equivalent to `missingsmallest(isless)`

bkamins: I would also add a comment about usage with sorting

src/Missings.jl
bkamins: as discussed earlier - let us start with the simplest possible definition that does not try to do optimizations.

src/Missings.jl
alonsoC1s: Sorry, I think I dont' understand the comment

bkamins: This is docstring of `missingsmallest(f)` signature (see the top of the string), but you are showing example of `missingsless(-Inf, missing)` which is a signature taking two positional arguments.

src/Missings.jl
```diff
@@ -514,4 +514,73 @@ julia> emptymissing(first)([1], 2)
...
+`missing` is the smallest possible value, and all other non-`missing` values are
+compared according to `f`.
+
+The traditional `isless` function modified to treat `missing` as the smallest
+value can be accessed by using the 2-argument `missingsmallest(x, y)` function,
+see [`missingsmallest`](@ref).
+
+# Examples
+```
+julia> ismuchless = missingsmallest((x, y) -> isless(x, 10*y))
```
LilithHafner, line 537: `ismuchless` is not a valid comparison function for `sort!`, I'd rather use an example that is a valid comparison function for `sort!`.

README.md
```diff
@@ -28,6 +28,8 @@ This package provides additional functionality for working with `missing` values
 - `Missings.replace` to wrap a collection in a (possibly indexable) iterator replacing `missing` with another value
 - `Missings.fail` to wrap a collection in a (possibly indexable) iterator throwing an error if `missing` is encountered
 - `skipmissings` to loop through a collection of iterators excluding indices where any iterators are `missing`
+- `missingsmallest(f)` to create a partial order function that treats `missing` as the smallest value and otherwise behaves like `f`
+- `missingsmallest` the standard `isless` function modified to treat `missing` as the smallest value rather than the largest one
```
nalimilan, line 32:
```suggestion
- `missingsmallest`: the standard `isless` function modified to treat `missing` as the smallest value rather than the largest one
```

src/Missings.jl
```diff
@@ -514,4 +514,73 @@ julia> emptymissing(first)([1], 2)
...
-end # module
+# Only for internal use. Allows dispatch over anonymous functions.
+struct MissingSmallest{T}
+    lt::T
+end
+
+"""
+    missingsmallest(f)
+
+Creates a function of two arguments `x` and `y` that tests whether `x` is less
```
nalimilan, line 525:
```suggestion
Return a function of two arguments `x` and `y` that tests whether `x` is less
```

src/Missings.jl
```diff
@@ -514,4 +514,73 @@ julia> emptymissing(first)([1], 2)
...
+than `y` such that `missing` is always less than the other argument. In other 
+words, returns a modified version of the partial order function `f` such that
```
nalimilan, line 527:
```suggestion
words, return a modified version of the partial order function `f` such that
```

src/Missings.jl
```diff
@@ -514,4 +514,73 @@ julia> emptymissing(first)([1], 2)
...
+`missing` is the smallest possible value, and all other non-`missing` values are
+compared according to `f`.
+
+The traditional `isless` function modified to treat `missing` as the smallest
+value can be accessed by using the 2-argument `missingsmallest(x, y)` function,
+see [`missingsmallest`](@ref).
```
nalimilan, line 533:
```suggestion
The behavior of the standard `isless` function modified to treat `missing` as the smallest
value can be obtained by calling the 2-argument `missingsmallest(x, y)` function. This
is equivalent to `missingsmallest(isless)(x, y)`.
```
(The link points to the same docstring.)

src/Missings.jl
```diff
@@ -514,4 +514,73 @@ julia> emptymissing(first)([1], 2)
...
+false
+```
+"""
+missingsmallest(f) = MissingSmallest(f)
+
+"""
+    missingsmallest(x, y)
+
+The standard partial order `isless` modified so that `missing` is always the
+smallest possible value. The expected behaviour is the following:
```
nalimilan, line 551:
```suggestion
smallest possible value:
```

src/Missings.jl
```diff
@@ -514,4 +514,73 @@ julia> emptymissing(first)([1], 2)
...
+- If neither argument is `missing`, the function behaves exactly as `isless`.
+- If `x` is `missing` the result will be `true` regardless of the value of `y`.
+- If `y` is `missing` the result will be `false` regardless of the value of `x`.
+
+See also [`missingsmallest`](@ref), the 1-argument function which takes a
```
nalimilan, line 556:
```suggestion
See also the 1-argument method which takes a
```

src/Missings.jl
```diff
@@ -514,4 +514,73 @@ julia> emptymissing(first)([1], 2)
...
+partial ordering function (like `isless`) and modifies it to treat `missing` as
+explained above. These functions can be used together with sorting functions
+such that the first elements of the newly sorted Array are placed first.
```
nalimilan, line 559:
```suggestion
so that missing values are sorted first. This is useful in particular so that
when sorting in reverse order missing values appear at the end.
```

src/Missings.jl
```diff
@@ -514,4 +514,73 @@ julia> emptymissing(first)([1], 2)
...
+
+# Examples
+```
+julia> v = [missing, 10, missing, 1, 2]
+julia> sort(v, lt=missingsmallest)
```
nalimilan, line 564:
```suggestion
julia> sort(v, lt=missingsmallest)

```

src/Missings.jl
```diff
@@ -514,4 +514,73 @@ julia> emptymissing(first)([1], 2)
...
+- If neither argument is `missing`, the function behaves exactly as `isless`.
+- If `x` is `missing` the result will be `true` regardless of the value of `y`.
+- If `y` is `missing` the result will be `false` regardless of the value of `x`.
+
+See also [`missingsmallest`](@ref), the 1-argument function which takes a
+partial ordering function (like `isless`) and modifies it to treat `missing` as
+explained above. These functions can be used together with sorting functions
+such that the first elements of the newly sorted Array are placed first.
+
+# Examples
```
nalimilan, line 561: Also show an example with `rev=true`, as it's the main motivation for this feature?

alonsoC1s:
> `ismuchless` is not a valid comparison function for `sort!`, I'd rather use an example that is a valid comparison function for `sort!`. 

Sorry, I don't see why `ismuchless` is not a valid comparison function. It works like this when you use it with `sort`
```
julia> v = [missing, 10, missing, 1, 2]
julia> sort(v, lt=ismuchless)
5-element Vector{Union{Missing, Int64}}:
   missing
   missing
  2
  1
 10
```

The 2 was placed before 1, but that makes intuitive sense right?

src/Missings.jl
alonsoC1s: Sorry, I don't see why ismuchless is not a valid comparison function. It works like this when you use it with sort

julia> v = [missing, 10, missing, 1, 2]
julia> sort(v, lt=ismuchless)
5-element Vector{Union{Missing, Int64}}:
   missing
   missing
  2
  1
 10
The 2 was placed before 1, but that makes intuitive sense right?

LilithHafner: In Julia 1.9

```
julia> sort(1:1000, lt=(x, y) -> isless(x, 10*y))
1000-element Vector{Int64}:
 1000
  999
  998
  997
  996
  995
    ⋮
    5
    4
    3
    2
    1
```

In Julia 1.6

```
julia> sort(collect(1:10000), lt=(x, y) -> isless(x, 10*y))

signal (11): Segmentation fault: 11
in expression starting at REPL[6]:1
getindex at ./array.jl:805 [inlined]
partition! at ./sort.jl:554
sort! at ./sort.jl:570
sort! at ./sort.jl:659 [inlined]
#sort!#8 at ./sort.jl:719 [inlined]
sort!##kw at ./sort.jl:707 [inlined]
#sort#9 at ./sort.jl:766 [inlined]
sort##kw at ./sort.jl:766
jl_apply_generic at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
do_call at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
eval_body at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
jl_interpret_toplevel_thunk at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
jl_toplevel_eval_flex at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
jl_toplevel_eval_flex at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
jl_toplevel_eval_in at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
eval at ./boot.jl:360 [inlined]
eval_user_input at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:139
repl_backend_loop at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:200
start_repl_backend at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:185
#run_repl#42 at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:317
run_repl at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:305
jl_apply_generic at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
#881 at ./client.jl:387
jfptr_YY.881_31907 at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/sys.dylib (unknown line)
jl_apply_generic at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
jl_f__call_latest at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
#invokelatest#2 at ./essentials.jl:708 [inlined]
invokelatest at ./essentials.jl:706 [inlined]
run_main_repl at ./client.jl:372
exec_options at ./client.jl:302
_start at ./client.jl:485
jfptr__start_43472 at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/sys.dylib (unknown line)
jl_apply_generic at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
true_main at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
repl_entrypoint at /Users/x/.julia/juliaup/julia-1.6.7+0.x64.apple.darwin14/lib/julia/libjulia-internal.1.6.dylib (unknown line)
Allocations: 4050783 (Pool: 4049777; Big: 1006); GC: 5
```

An excerpt from the 1.11 docstring: lt must be "transitive in equivalence: !lt(x, y) && !lt(y, x) and !lt(y, z) && !lt(z, y) together imply !lt(x, z) && !lt(z, x). In words: if x and y are equivalent and y and are equivalent then x and z must be equivalent."

src/Missings.jl
nalimilan: I'm not sure it's worth adding a reference, as both docstrings will be grouped in the manual and in the REPL.

bkamins: Yes - I think this cross reference is not needed in this case.

src/Missings.jl
bkamins: Yes - I would recommend some natural example, e.g. comparing strings by their length (and if they have the same length they are equivalent).

test/runtests.jl
```diff
@@ -257,4 +257,22 @@ struct CubeRooter end
     @test emptymissing(fun)(3, 1, c=2) == (1, 2)
 end
 
+@testset "missingsmallest" begin
+    @test missingsmallest(missing, Inf) == true
+    @test missingsmallest(-Inf, missing) == false
+    @test missingsmallest(missing, missing) == false
+    @test missingsmallest(3, 4) == true
+    @test missingsmallest(-Inf, Inf) == true
```
bkamins, line 265: maybe also add tests with something else than numbers, e.g. strings?

test/runtests.jl
```diff
@@ -257,4 +257,22 @@ struct CubeRooter end
...
 end
 
+@testset "missingsmallest" begin
+    @test missingsmallest(missing, Inf) == true
+    @test missingsmallest(-Inf, missing) == false
+    @test missingsmallest(missing, missing) == false
+    @test missingsmallest(3, 4) == true
+    @test missingsmallest(-Inf, Inf) == true 
+
+    ≪(x, y) = isless(10*x, y) # "Much greater than" function
```
bkamins, line 267: This test should be replaced as @LilithHafner suggested.

src/Missings.jl
alonsoC1s: Wow, I had no idea that could cause such a serious error 🫠 Today I learned
I will remove the example entirely then

src/Missings.jl
```diff
@@ -514,4 +514,88 @@ julia> emptymissing(first)([1], 2)
...
+```
+julia> lengthmissing = passmissing(length)
+julia> isshorter = missingsmallest((s1, s2) -> isless(lengthmissing(s1), lengthmissing(s2)))
+julia> isshorter("short", "longstring")
+true
+
+julia> isshorter("longstring", "short")
+false
+
+julia> isshorter("", missing) # Is shorter than length 0?
```
bkamins, line 545: I am not clear what the comment means here

bkamins (approved):

src/Missings.jl
```diff
@@ -514,4 +514,88 @@ julia> emptymissing(first)([1], 2)
...
+compared according to `f`.
+
+The behavior of the standard `isless` function modified to treat `missing` as
+the smallest value can be obtained by calling the 2-argument `missingsmallest(x,
+y)` function. This is equivalent to `missingsmallest(isless)(x, y)`.
+
+# Examples
+```
+julia> lengthmissing = passmissing(length)
+julia> isshorter = missingsmallest((s1, s2) -> isless(lengthmissing(s1), lengthmissing(s2)))
```
LilithHafner, line 538:
```suggestion
julia> isshorter = missingsmallest((s1, s2) -> isless(length(s1), length(s2)))
```
`missingsmallest` will take care of missings, we don't also need to `passmissing`.

src/Missings.jl
```diff
@@ -514,4 +514,88 @@ julia> emptymissing(first)([1], 2)
...
+missingsmallest(f) = MissingSmallest(f)
+
+"""
+    missingsmallest(x, y)
+
+The standard partial order `isless` modified so that `missing` is always the
+smallest possible value:
+- If neither argument is `missing`, the function behaves exactly as `isless`.
+- If `x` is `missing` the result will be `true` regardless of the value of `y`.
+- If `y` is `missing` the result will be `false` regardless of the value of `x`.
```
LilithHafner, line 558:
```suggestion
- If `y` is `missing` the result will be `false` regardless of the value of `x`.
- If `x` is `missing` the result will be `true` unless `y` is `missing`.
```

src/Missings.jl
```diff
@@ -514,4 +514,88 @@ julia> emptymissing(first)([1], 2)
...
+
+See also the 1-argument method which takes a partial ordering function (like
+`isless`) and modifies it to treat `missing` as explained above. These functions
+can be used together with sorting functions so that missing values are sorted
+first. This is useful in particular so that when sorting in reverse order
+missing values appear at the end.
+
+# Examples
+```
```
LilithHafner, line 567:
```suggestion
```jldoctest
```

src/Missings.jl
```diff
@@ -514,4 +514,88 @@ julia> emptymissing(first)([1], 2)
...
+   missing
+
+julia> missingsmallest(missing, Inf)
+true
+
+julia> missingsmallest(-Inf, missing)
+false
+
+julia> missingsmallest(missing, missing)
+true
```
LilithHafner, line 594:
```suggestion
false
```

And adding `jldoctest` above should catch this sort of error automatically.

src/Missings.jl
bkamins: Ah, the reason is that we have `ismissing(y) ? false : ismissing(x) ? true : ms.lt(x, y)` so if `x` or `y` is `missing` then `length` will not be called. Right?

LilithHafner: yep!

bkamins:
@alonsoC1s - is it clear what should be corrected?

alonsoC1s:
I believe so. I'll get to work on the latests comments :)

LilithHafner (approved):

LilithHafner:
Nightly failures are unrelated.

LilithHafner:
My inclination is to make the documentation of `missingsmallest`  explicit about `isless` serving that role rather than adding `missinglargest` because I'm worried that the existence of `missinglargest` could be confusing, but I could see it going either way.

LilithHafner:
@bkamins, should this be merged?

bkamins:
I approved it already some time ago. We are waiting for @nalimilan before merging.

src/Missings.jl
```diff
@@ -514,4 +514,85 @@ julia> emptymissing(first)([1], 2)
...
+
+5-element Vector{Union{Missing, Int64}}:
+   missing
+   missing
+  1
+  2
+ 10
+
+julia> sort(v, lt=missingsmallest, rev=true)
+
```
nalimilan, line 577:
```suggestion
```

nalimilan (commented):
Sorry. I've found a pair of remaining issues in doctests (as unfortunately we don't test them since we don't build a manual).

pdeffebach:
@LilithHafner Bumping this. The final changes seem pretty small before merging.

LilithHafner:
Sorry about the delay. I recommend accepting @nalimilan's suggestions and then then merging into main if CI still passes.

I don't have commit bit here. cc @bkamins, @nalimilan.

pdeffebach:
Ah sorry! I thought you were the one making the changes! @alonsoC1s can you make the final changes?

bkamins:
@alonsoC1s - when you finalize the PR can you also update the tests so that they pass both on the current release and on nightly? (there is an issue I think that depending on Julia version a different exception type is thrown in some cases). Thank you!

bkamins:
I have commited the suggestions.

alonsoC1s:
@bkamins sure, I'll be happy to. This week is a bit busy, but I will wrap everything up as soon as I can get some free time

alonsoC1s:
@bkamins I'm taking a look at the reasons for failure on different Julia versions and I think you are right, different versions throw different Errors. What is the best practice for dealing with this version-specific behaviour?

bkamins:
I typically test for an abstract error that is a supertype of the errors thrown.
Alternative is to write version-specific tests, but this is more cumbersome.

alonsoC1s:
Agreed, I'll test for more general Exceptions

bkamins:
@nalimilan - OK to merge?

src/Missings.jl
```diff
@@ -514,4 +514,85 @@ julia> emptymissing(first)([1], 2)
...
+See also the 1-argument method which takes a partial ordering function (like
+`isless`) and modifies it to treat `missing` as explained above. These functions
+can be used together with sorting functions so that missing values are sorted
+first. This is useful in particular so that when sorting in reverse order
+missing values appear at the end.
+
+# Examples
+```jldoctest
+julia> sort(v, lt=missingsmallest)
+
```
nalimilan, line 569:
```suggestion
```

nalimilan (approved):
Thanks!

bkamins:
Thank you!

PR #149, opened 2025-05-22
Create dependabot.yml and update action versions

abhro:


nalimilan (approved):
Based on https://github.com/JuliaLang/Example.jl/tree/master/.github I would say it's OK.

PR #152, opened 2026-02-07
Use 'min' Julia version in ci.yml

abhro:
Make the CI test on lowest supported version as stated in Project.toml. Should also serve as an indicator if the minimum Julia version needs to be bumped

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,20 +40,3 @@ jobs:
         with:
           files: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
-#  docs:
-#    name: Documentation
-#    runs-on: ubuntu-latest
-#    steps:
-#      - uses: actions/checkout@v6
-#      - uses: julia-actions/setup-julia@v2
-#        with:
-#          version: '1'
-#      - run: |
-#          julia --project=docs -e '
-#            using Pkg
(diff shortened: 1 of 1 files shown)


abhro:
gentle bump

bkamins:
I think it would be good if you mention someone who knows these configuation files and could competently review it. Thank you!

abhro:
Hmm. I'm not sure who the current maintainers of this package are. I'll dig around.

abhro:
From the blame, it seems @quinnj was most involved in CI setup. Requesting review.

bkamins:
The issue is about competences. E.g. I could review implementation changes, but these are CI changes and I do not know current standards in this domain.

abhro:
Requesting review from @quinnj

DilumAluthge (commented):
I personally prefer the old branches syntax, could you revert back to that?

Apart from that LGTM.

.github/workflows/ci.yml
```diff
@@ -4,7 +4,7 @@ on:
     branches:
       - main
   push:
-    branches:
+    branches:
```
DilumAluthge:
```suggestion
    branches:
```

DilumAluthge (approved):

DilumAluthge (approved):

